PR #158, opened 2022-07-04
New CD deployment pipeline & new infra docs & /version endpoint & validator SSL support 

jsign:
This PR contains the new CD setup, which is part of a bigger change in our new infrastructure.

Some notes:
- I also added support to add SSL certs to the validator if the endpoints should be served in `HTTPS` or require origin certificates being behind an SSL offloading service.
- The new GH Action setup will be explained via PR comments.
- A new `/version` REST endpoint was added which shows git summary of the binary being executed; this will be useful for reasons I'll describe in PR comments.
- To understand the new infra setup as a whole, I wrote a document that explains the big picture, once-setups, and the steps to prepare any new VM for the environment. Read it [here](https://www.notion.so/textile/SQLite-Validator-infrastructure-0dddb0ceae6d4e93aa56df7df4658993).
- The `staging.tableland.network` DNS is pointing to this new `staging` VM, and working.
- I decommissioned the previous `staging ` in k8 + Postgres gcloud VM.


.github/workflows/deploy.yml
```diff
@@ -12,98 +12,50 @@ on:
     types:
```
jsign, line 12: Maybe be easier to see without diffs.
Also, I'd recommend reading the linked Notion doc to understand the infrastructure setup before reading this.

.github/workflows/deploy.yml
```diff
@@ -12,98 +12,50 @@ on:
...
-        uses: actions/cache@v2
-        with:
-          path: /tmp/.buildx-cache
-          key: ${{ runner.os }}-buildx-${{ github.sha }}
-          restore-keys: |
-            ${{ runner.os }}-buildx-
+          # This is the official way to setup secret values that are multi-line, such as an SSH private key.
+          echo 'vm_ssh_key<<EOF' >> $GITHUB_ENV
+          echo "${{ secrets.VM_SSH_KEY }}" >> $GITHUB_ENV
+          echo 'EOF' >> $GITHUB_ENV
```
jsign, line 38: In this `Prepare` step, we setup two new outputs:
- `vm_host` which will have the VM IP that we'll connect via SSH
- `vm_ssh_key` which contains the SSH keys to connect to the VM.

Note that `vm_host` is conditional depending on the usual commit substring `[<env>deploy]` idea or branch.
The `vm_ssh_key` is independent since all the VMs are configured to accept connections from a single SSH key that we use for GH actions.

Note that `vm_ssh_key` is a multiline value, so we need to do what's done in L35 to set up properly.

.github/workflows/deploy.yml
```diff
@@ -12,98 +12,50 @@ on:
...
+            cd go-tableland
+            git fetch && git checkout -f ${{ github.sha }}
+            cd docker
+            HEALTHBOT_ENABLED=true make ${{ steps.prep.outputs.deployment }}-up
+
+            # Wait to start and double-check we're running the expected version.
+            # This helps having a green check in the GH Action be high-signal that everything is fine.
+            sleep 5;
+            for i in 1 2 3 4; do [ $(curl --insecure --silent https://127.0.0.1/version | jq .GitCommit) = "\"${{ steps.prep.outputs.sha_short}}\"" ] && break || (if [ $i = 4 ]; then exit -1; else sleep 5; fi ) done 
+            echo "All healthy!"
```
jsign, line 60: This is the step that does the deployment, which leverages the `vm_host` and `vm_ssh_key` prepared above.

At first, I thought that we could simply `./deploy.sh` some `deploy.sh` file that we prepared on each VM, but I actually think it's better to put the script explicitly in the `script:` section, since that would allow the script to be versioned and also transparent.

The steps are pretty self describing, but some notes:
- The SSH keys of the action will connect to the `validator` user in the VM, so by default in its home folder where `go-tableland` lives.
- We `cd` there, fetch and checkout the sha that we want to deploy.
- We leverage the `steps.prep.outputs.deployment` to execute `make <env>-up` with healthbot enabled.

The code in `L56-L60` intention is twofold:
- Being paranoid about docker-compose and its caching mechanism; I've seen weird things happening in the past.
- I wanted that, whenever we see a "green check" in this GH action, we are 100% sure the validator *is* running and the code we expected.

For that I leverage the new `/version` REST endpoint that I created too. We `curl` there, get the `GitCommit`, and compare it with the SHA that we theoretically deployed. These values **must** match. If that's the case, we allow the action to finish with a green check. [I added some waiting time logic just to give it some time to spin up].

.github/workflows/deploy.yml
```diff
@@ -12,98 +12,50 @@ on:
...
 
-      - name: Deploy to GKE
-        working-directory: ./k8
-        run: |
-          gcloud container clusters get-credentials $GKE_CLUSTER --region $GKE_REGION --project ${{ secrets.GKE_PROJECT }}
-          export API_IMAGE=$CONTAINER_REGISTRY/${{ secrets.GKE_PROJECT }}/textile/tableland/api:sha-${{ steps.prep.outputs.sha_short }}
-          export HEALTHBOT_IMAGE=$CONTAINER_REGISTRY/${{ secrets.GKE_PROJECT }}/textile/tableland/healthbot:sha-${{ steps.prep.outputs.sha_short }}
-          DEPLOYMENT=${{ steps.prep.outputs.deployment }} make deploy
```
jsign, line 109: As a last comment about this action; I think we should eventually still create _public_ Docker images.
Maybe only in creating releases in the repo or similar; I think some people would value that.

cmd/api/main.go
```diff
@@ -108,15 +110,51 @@ func main() {
...
+			if err := server.ListenAndServe(); err != nil {
+				if err == http.ErrServerClosed {
+					log.Info().Msg("http serve gracefully closed")
+					return
+				}
+				log.Fatal().
+					Err(err).
+					Str("port", config.HTTP.Port).
+					Msg("could not start http server")
 			}
```
jsign, line 157: TL;DR is: if TLS cert and keys were provided, we default to listening on 443 using it.

docker/Makefile
```diff
@@ -15,9 +15,11 @@ staging-up:
 	PLATFORM=$(shell uname -m) \
 	ENVIROMENT=deployed/staging \
 	COMPOSE_DOCKER_CLI_BUILD=1 \
-	docker compose -f docker-compose.yml -f deployed/docker-compose.observability.yml $(HEALTHBOT_OVERRIDE) up --build --remove-orphans
+	docker compose -f docker-compose.yml -f deployed/docker-compose.observability.yml $(HEALTHBOT_OVERRIDE) up --build --remove-orphans --detach
```
jsign, line 18: I added the `--detach` since we want `make <env>-up` to finish and not block forever, so the action can actually finish.

docker/Makefile
```diff
@@ -15,9 +15,11 @@ staging-up:
...
 .PHONY: staging-up
 
 staging-down:
+	PLATFORM=$(shell uname -m) \
+	ENVIROMENT=deployed/staging \
 	docker compose -f docker-compose.yml -f deployed/docker-compose.observability.yml $(HEALTHBOT_OVERRIDE)  down
```
jsign, line 24: This was missing.

docker/deployed/docker-compose.observability.yml
```diff
@@ -20,4 +20,4 @@ services:
       - GF_USERS_ALLOW_SIGN_UP=false
     restart: unless-stopped
     ports:
-      - "127.0.0.1:3000:3000"
+      - "0.0.0.0:3000:3000"
```
jsign, line 23: Bind to `0.0.0.0` so it's actually listening for external requests (if firewall config allows).

docker/deployed/staging/api/config.json
```diff
@@ -31,7 +33,7 @@
       "Registry": {
         "EthEndpoint": "wss://opt-kovan.g.alchemy.com/v2/${VALIDATOR_ALCHEMY_OPTIMISM_KOVAN_API_KEY}",
         "ChainID": 69,
-        "ContractAddress": "0xEb64cA7F57729ac64a6eD3dc8F7bb03b570baeAD"
+        "ContractAddress": "0x2b51D72F210A96B6aB8EfB02dF8802363dcC83Dd"
```
jsign, line 36: The new staging contract.

docker/docker-compose.yml
```diff
@@ -10,7 +10,8 @@ services:
     env_file:
       - ${PWD}/${ENVIROMENT}/api/.env_validator
     ports:
-      - "127.0.0.1:8080:8080" 
+      - "0.0.0.0:8080:8080" 
+      - "0.0.0.0:443:443"
```
jsign, line 14: We open both 8080 and 443. The validator will be serving only in one depending on its TLS cert config. But we do both bindings in the compose file so both are fine to be reached.

internal/router/controllers/user.go
```diff
@@ -331,10 +331,10 @@ func (c *UserController) runReadRequest(
 			Err(err)
 
 		_ = json.NewEncoder(rw).Encode(errors.ServiceError{Message: err.Error()})
-		return sqlstore.UserRows{}, false
+		return nil, false
 	}
 
-	rows, ok := res.Result.(sqlstore.UserRows)
+	rows, ok := res.Result.(*sqlstore.UserRows)
```
jsign, line 337: This is a bug fix that I caught.

buildinfo/buildinfo.go
```diff
@@ -17,15 +15,25 @@ var (
...
+	GitCommit  string
+	GitBranch  string
+	GitState   string
+	GitSummary string
+	BuildDate  string
+	Version    string
+}
+
+// GetSummary returns a JSON string with a summary of git information.
+func GetSummary() (Summary, error) {
```
brunocalza, line 29: Maybe I'm missing something or this is a leftover of another logic. Does returning `error` make sense, if the code always returns `nil`?

brunocalza (approved):
Really nice. 

Did a comment here about the returning of `error` in `GetSummary ` and comment in the document about merging to `main` deploying to `staging` as well (not only `testnet`).

So good to see all the k8 configs being deleted. I think we are heading to a much simpler deployment setup.

Also, I think it's worth it (on my part) to spend some time setting up a new validator to see if I catch anything missing.

buildinfo/buildinfo.go
jsign: Probably a left over, I'll remove it :+1:

PR #161, opened 2022-07-06
Updates the parser

brunocalza:
- Contains table alias bug fix
- Contains support for identifiers wrapped by double quotes, backticks or square brackets

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -15,7 +15,7 @@ require (
 	github.com/sethvargo/go-limiter v0.7.2
 	github.com/spruceid/siwe-go v0.2.0
 	github.com/stretchr/testify v1.7.4
-	github.com/tablelandnetwork/sqlparser v0.0.0-20220627210906-4f88b6ff9726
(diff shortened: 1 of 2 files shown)


go.mod
```diff
@@ -15,7 +15,7 @@ require (
 	github.com/sethvargo/go-limiter v0.7.2
 	github.com/spruceid/siwe-go v0.2.0
 	github.com/stretchr/testify v1.7.4
-	github.com/tablelandnetwork/sqlparser v0.0.0-20220627210906-4f88b6ff9726
+	github.com/tablelandnetwork/sqlparser v0.0.0-20220706201905-0ca67a566c4e
```
jsign, line 18: Nit: maybe it's better to target `651da656fce841b8b9ca00fc409cb2e27ab8c09e` which is the `main` commit and not the PR one? (mostly to reference always `main` commits; the code is the same)

jsign (approved):
:100:

go.mod
jsign: Missed comment, done [here](https://github.com/tablelandnetwork/go-tableland/pull/162#discussion_r915263626).

PR #162, opened 2022-07-06
parser: disallow sqlite_ prefix in create tables

jsign:
We were missing disallowing `sqlite_` prefixes in table names. We had tests to exactly test this behavior, but in the main configuration we missed adding it.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -15,7 +15,7 @@ require (
 	github.com/sethvargo/go-limiter v0.7.2
 	github.com/spruceid/siwe-go v0.2.0
 	github.com/stretchr/testify v1.7.4
-	github.com/tablelandnetwork/sqlparser v0.0.0-20220706201905-0ca67a566c4e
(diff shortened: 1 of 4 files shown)


go.mod
```diff
@@ -15,7 +15,7 @@ require (
 	github.com/sethvargo/go-limiter v0.7.2
 	github.com/spruceid/siwe-go v0.2.0
 	github.com/stretchr/testify v1.7.4
-	github.com/tablelandnetwork/sqlparser v0.0.0-20220706201905-0ca67a566c4e
+	github.com/tablelandnetwork/sqlparser v0.0.0-20220706202446-651da656fce8
```
jsign, line 18: Refer to `main` commit.

internal/router/controllers/user_test.go
```diff
@@ -161,13 +161,13 @@ func (rm *runnerMock) RunReadQuery(
 		str := "foo"
```
jsign, line 161: Test fixing.

brunocalza (approved):

PR #164, opened 2022-07-07
Detect and return raw json data

asutula:


pkg/sqlstore/store_user.go
```diff
@@ -17,6 +19,53 @@ type UserRows struct {
...
+	}
+	return u.otherValue
+}
+
+// Scan implements Scan.
+func (u *UserData) Scan(src interface{}) error {
+	u.jsonValue = nil
+	u.otherValue = nil
+	switch src := src.(type) {
+	case string:
```
carsonfarmer, line 41: In this case, the string might still contain a valid JSON value that isn't an object or array, right? So an int will end up being returned as a string? That is probably ok, but just wanted to confirm.

sanderpick:
@asutula looks like your proposal has been sufficiently consented to. Should we take this out of draft?

pkg/eventprocessor/impl/eventprocessor_test.go
```diff
@@ -293,17 +293,20 @@ type contractCalls struct {
...
-type contractTransferFromSender func(controller common.Address) common.Hash
-type checkReceipts func(*testing.T, ...eventprocessor.Receipt) func() bool
+type (
+	dbReader                    func(string) []int64
+	contractRunSQLBlockSender   func([]string) []common.Hash
+	contractCreateTableSender   func(string) common.Hash
+	contractSetControllerSender func(controller common.Address) common.Hash
+	contractTransferFromSender  func(controller common.Address) common.Hash
+	checkReceipts               func(*testing.T, ...eventprocessor.Receipt) func() bool
+)
```
asutula, line 303: Started using `gofumpt` formatter instead of `gopls`. Apparently in includes all `gopls` but is stricter and does extra stuff like this. So far I like it.

pkg/eventprocessor/impl/eventprocessor_test.go
```diff
@@ -419,6 +422,7 @@ func (acl *aclMock) CheckPrivileges(
 	tx *sql.Tx,
 	controller common.Address,
 	id tableland.TableID,
-	op tableland.Operation) (bool, error) {
+	op tableland.Operation,
+) (bool, error) {
```
asutula, line 426: `gofumpt` strikes again. I like this too, this is actually how I would normally format this manually.

pkg/sqlstore/impl/user/rowstojson.go
```diff
@@ -40,7 +40,7 @@ func getRowsData(rows *sql.Rows, numColumns int) ([][]interface{}, error) {
 	for rows.Next() {
 		scanArgs := make([]interface{}, numColumns)
 		for i := range scanArgs {
-			scanArgs[i] = new(interface{})
+			scanArgs[i] = &sqlstore.UserValue{}
```
asutula, line 43: Always scan into `UserValue`

pkg/sqlstore/store_user.go
```diff
@@ -17,6 +19,49 @@ type UserRows struct {
...
+	if u.jsonValue != nil {
+		return u.jsonValue
+	}
+	return u.otherValue
+}
+
+// Scan implements Scan.
+func (u *UserValue) Scan(src interface{}) error {
+	u.jsonValue = nil
+	u.otherValue = nil
```
asutula, line 39: Clearing these out for safety.

pkg/sqlstore/store_user.go
```diff
@@ -17,6 +19,49 @@ type UserRows struct {
...
+	case string:
+		trimmed := strings.TrimLeft(src, " ")
+		if (strings.HasPrefix(trimmed, "{") || strings.HasPrefix(trimmed, "[")) && json.Valid([]byte(src)) {
+			u.jsonValue = []byte(src)
+		} else {
+			u.otherValue = src
+		}
+	case []byte:
+		tmp := src
+		u.otherValue = tmp
```
asutula, line 50: Added this because of a dangerous sounding comment in the [`Scanner` declaration](https://pkg.go.dev/database/sql#Scanner) that 

> // Reference types such as []byte are only valid until the next call to Scan
> // and should not be retained. Their underlying memory is owned by the driver.
> // If retention is necessary, copy their values before the next call to Scan.

pkg/sqlstore/store_user.go
```diff
@@ -17,6 +19,49 @@ type UserRows struct {
...
+	return nil
+}
+
+// MarshalJSON implements MarshalJSON.
+func (u *UserValue) MarshalJSON() ([]byte, error) {
+	if u.jsonValue != nil {
+		return u.jsonValue, nil
+	}
+	return json.Marshal(u.otherValue)
+}
```
asutula, line 63: When marshaling to json, there is nothing to do for `jsonValue` nice we already identified it as valid json.

pkg/sqlstore/store_user.go
jsign: Good call.

Important: `tmp := src` isn't the kind of copy `Scanner` is expecting. This is a shallow copy of the reference of the underlying slice, so `tmp` will still share the same bytes as `src`.

We should:
```
tmp := make([]byte, len(src))
copy(tmp, src)
```

pkg/eventprocessor/impl/eventprocessor_test.go
```diff
@@ -374,7 +377,7 @@ func setup(t *testing.T) (
 		queryRes := res.(*sqlstore.UserRows)
 		ret := make([]int64, len(queryRes.Rows))
 		for i := range queryRes.Rows {
-			ret[i] = (*queryRes.Rows[i][0].(*interface{})).(int64)
+			ret[i] = (*queryRes.Rows[i][0].(*sqlstore.UserValue)).Value().(int64)
```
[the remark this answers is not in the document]
jsign: I think a similar change should be done [here](https://github.com/tablelandnetwork/go-tableland/blob/89637eb0cd23fe1fa1ca12a193ac37f94c3102d3/internal/router/controllers/user.go#L146).
Maybe can be double-checked with a test.

jsign (approved):
LGTM, I just left a comment with the `Scanner` deep copy that the API mentions is necessary, and an extra place that I think we need to change type conversion.

pkg/eventprocessor/impl/eventprocessor_test.go
asutula: Good question. Looking. Indeed it can https://golangci-lint.run/usage/linters/#gofumpt

pkg/eventprocessor/impl/eventprocessor_test.go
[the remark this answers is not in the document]
asutula: Ah yea good catch. The current tests pass because the mocked data is returning `string`, not `UserData`. Hmm, I'll look into updating the mock data and that line or get to reading and considering your other comment about not using `UserData`.

pkg/sqlstore/impl/user/rowstojson.go
asutula: I think this is a good idea and avoids having to refactor tests, potentially miss some other required changes being hidden by mocked test data. Will try this out.

jsign: Cool, remember to include the extra copy logic that you found from the `Scanner` interface below to handle `[]byte` since that's the correct thing since we retain `[]bytes` append in L48. So prob mostly "moving" the logic that already exists into a `switch` between lines L47 and L48 iterating in `scanArgs...` or something like that.

asutula: Hey @jsign I'm spinning my wheels a bit trying to figure out the type of the value pointed to by a `*interface{}`. I'm tempted to use reflection api, but is there any way to avoid that?

jsign: You can do the same as I mentioned in the other comment [here](https://github.com/tablelandnetwork/go-tableland/blob/89637eb0cd23fe1fa1ca12a193ac37f94c3102d3/internal/router/controllers/user.go#L146)?

So prob: `switch (*xxx).(*interface{})).(type) {` ?

asutula: oh yea, thanks. all this indirecting and `interface{}` stuff can get confusing.

pkg/sqlstore/impl/user/rowstojson.go
```diff
@@ -45,6 +47,19 @@ func getRowsData(rows *sql.Rows, numColumns int) ([][]interface{}, error) {
...
+			switch src := (*scanArg.(*interface{})).(type) {
+			case string:
+				trimmed := strings.TrimLeft(src, " ")
+				if (strings.HasPrefix(trimmed, "{") || strings.HasPrefix(trimmed, "[")) && json.Valid([]byte(trimmed)) {
+					scanArgs[i] = []byte(trimmed)
+				}
+			case []byte:
+				tmp := make([]byte, len(src))
+				copy(tmp, src)
+				scanArgs[i] = tmp
```
asutula, line 60: Not clear to me this is doing anything? Like, if we didn't do this PR at all, `[]byte` values would be handled fine, so I'm not sure why we need to do this.

pkg/sqlstore/impl/user/rowstojson.go
```diff
@@ -45,6 +47,19 @@ func getRowsData(rows *sql.Rows, numColumns int) ([][]interface{}, error) {
...
+			}
+		}
 		rowsData = append(rowsData, scanArgs)
```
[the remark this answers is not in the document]
asutula: Just want to be sure this is right, then I revert all the unneeded changes in the PR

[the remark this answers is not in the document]
jsign: Yes, I think looks reasonable. I dunno if you have done tests with this solution, but if works I think it's simpler.

pkg/sqlstore/impl/user/rowstojson.go
jsign: This is not putting in `scanArgs[i]` the bytes returned by the driver, since the driver might reuse those bytes in the next row (override).
This is something that wasn't right before this PR.

Maybe `mattn/sqlite3` doesn't reuse the bytes for each row since that's just a constraint put at the interface level of `database/sql`, so maybe some drivers prefer to lean on the safe side.

jsign: I also not entirely sure how many people have really used `BLOB` columns to make this potential bug be noticed.

pkg/sqlstore/store_user.go
```diff
@@ -13,8 +15,62 @@ type UserColumn struct {
 
 // UserRows defines a row result.
 type UserRows struct {
-	Columns []UserColumn    `json:"columns"`
-	Rows    [][]interface{} `json:"rows"`
+	Columns []UserColumn   `json:"columns"`
+	Rows    [][]*UserValue `json:"rows"`
```
asutula, line 19: I realized that since we're always scanning into `UserValue`, this could be typed now, and that change bubbles up through lots of places giving us strong typing. No need for asserting type in some places now.

pkg/sqlstore/store_user.go
```diff
@@ -13,8 +15,62 @@ type UserColumn struct {
...
+		trimmed := strings.TrimLeft(src, " ")
+		if (strings.HasPrefix(trimmed, "{") || strings.HasPrefix(trimmed, "[")) && json.Valid([]byte(src)) {
+			u.jsonValue = []byte(src)
+		} else {
+			u.otherValue = src
+		}
+	case []byte:
+		tmp := make([]byte, len(src))
+		copy(tmp, src)
+		u.otherValue = tmp
```
asutula, line 51: This change is in here.

pkg/sqlstore/store_user.go
```diff
@@ -13,8 +15,62 @@ type UserColumn struct {
...
+
+// JSONUserValue creates a UserValue with the provided json.
+func JSONUserValue(v json.RawMessage) *UserValue {
+	return &UserValue{jsonValue: v}
+}
+
+// OtherUserValue creates a UserValue with the provided other value.
+func OtherUserValue(v interface{}) *UserValue {
+	return &UserValue{otherValue: v}
 }
```
asutula, line 74: Some helpers to create values for mock test data.

pkg/sqlstore/impl/user/rowstojson.go
```diff
@@ -35,18 +35,22 @@ func getColumnsData(rows *sql.Rows) ([]sqlstore.UserColumn, error) {
...
+			scanArgs[i] = vals[i]
 		}
 		if err := rows.Scan(scanArgs...); err != nil {
 			return nil, fmt.Errorf("scan row column: %s", err)
 		}
-		rowsData = append(rowsData, scanArgs)
+		rowsData = append(rowsData, vals)
 	}
-
 	return rowsData, nil
```
asutula, line 55: This is strongly typed now.

pkg/eventprocessor/impl/eventprocessor_test.go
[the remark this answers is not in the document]
asutula: Fixed via strong typing of rows with `UserValue`

pkg/sqlstore/store_user.go
asutula: Done.

pkg/sqlstore/impl/user/rowstojson.go
asutula: Hey @jsign give this another look over in the morning. I reverted to using `UserValue` and also realized that `UserRows.Rows` can now be `[][]*UserValue` which had a nice effect bubble up and eliminate places where we were asserting types from `interface{}`.

asutula:
Hey @jsign give this another look over in the morning. I reverted to using `UserValue` and also realized that `UserRows.Rows` can now be of type `[][]*UserValue` which had a nice effect bubble up and eliminate places where we were asserting types from `interface{}`.

PR #167, opened 2022-07-08
fix prefix in table row gateway query

sanderpick:
fixes panic: `interface conversion: interface {} is *interface {}, not **string`

afaict, conversion from `*interface {}` requires reflect to get at the underlying value type.

### DIFF
--- a/internal/router/controllers/user.go
+++ b/internal/router/controllers/user.go
@@ -5,7 +5,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
-	"reflect"
 	"strconv"
 
 	"github.com/gorilla/mux"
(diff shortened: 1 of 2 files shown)


internal/router/controllers/user.go
```diff
@@ -143,7 +144,7 @@ func (c *UserController) GetTableRow(rw http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	prefix := **prefixRow.Rows[0][0].(**string)
+	prefix := reflect.Indirect(reflect.ValueOf(prefixRow.Rows[0][0]))
```
jsign, line 147: Alternative: This can also be: `prefix := (*prefixRow.Rows[0][0].(*interface{})).(string)`.
The way to think about it is that scanned results from queries are `*interface{}`, if we dereference that we should find the underlying "expected" type.

jsign (approved):
:heavy_check_mark:

PR #168, opened 2022-07-08
docker/grafana: bind the data folder to allow alerting persistence

jsign:
Bind grafana `/data` folder as we do with Prometheus.

This is needed since Grafana still doesn't support alert provisioning as we do with the dashboard, so it can't be stateless. We'll need to create the alerts manually, and have that be preserved in the data folder.

Whenever Grafana supports alerts provisioning (which they mention it's the plan), we can have the same style of configuration as we do with data sources and dashboards, and make it stateless again.

Also fixing some typos and updating deps.


docker/Makefile
```diff
@@ -13,26 +13,30 @@ HEALTHBOT_OVERRIDE=$(shell if [ ${HEALTHBOT_ENABLED} = "true" ]; then echo "-f d
 
 staging-up:
 	PLATFORM=$(shell uname -m) \
-	ENVIROMENT=deployed/staging \
+	USER=$(shell id -u) \
```
jsign, line 16: Note that we set `USER` to the current user. 

This is needed since the Grafana docker image sets some user id that usually doesn't exist in the host. This creates problems with permissions in the data folder to save the state. 

Here we follow official advice of running the container with the user id that is currently running. In the case of our infra would be `validator` which has access in the new `grafana/data` folder.

brunocalza (approved):
Looks good!

PR #169, opened 2022-07-09
Add rpc endpoint validateWriteQuery

joewagner:
This is a first try at writing production go, so any feedback would be really appreciated.  Hopefully it satisfies the Validator's part of https://www.notion.so/textile/Add-enableTestnetRPCRelay-flag-in-JS-SDK-d6ccf9e045bb4c54a43bbb0b0e78483d

### DIFF
--- a/internal/tableland/impl/mesa.go
+++ b/internal/tableland/impl/mesa.go
@@ -58,12 +58,14 @@ func (t *TablelandMesa) ValidateWriteQuery(
 	if !ok || controller == "" {
 		return tableland.ValidateWriteQueryResponse{}, errors.New("no controller address found in context")
 	}
+
 	ctxChainID := ctx.Value(middlewares.ContextKeyChainID)
 	chainID, ok := ctxChainID.(tableland.ChainID)
 	if !ok {
 		return tableland.ValidateWriteQueryResponse{}, errors.New("no chain id found in context")
 	}
-	_, chainOk := t.chainStacks[chainID]
+
+	stack, chainOk := t.chainStacks[chainID]
 	if !chainOk {
 		return tableland.ValidateWriteQueryResponse{}, fmt.Errorf("chain id %d isn't supported in the validator", chainID)
 	}
@@ -74,6 +76,19 @@ func (t *TablelandMesa) ValidateWriteQuery(
 	}
 
 	tableID := mutatingStmts[0].GetTableID()
+
+	table, err := stack.Store.GetTable(ctx, tableID)
+
+	// if the tableID is not valid err will exist
+	if err != nil {
+		return tableland.ValidateWriteQueryResponse{}, fmt.Errorf("getting table: %s", err)
+	}
+	// if the prefix is wrong the statement is not valid
+	prefix := mutatingStmts[0].GetPrefix()
+	if table.Prefix != prefix {
+		return tableland.ValidateWriteQueryResponse{}, fmt.Errorf(
+			"table prefix doesn't match (exp %s, got %s)", table.Prefix, prefix)
+	}
 
 	response := tableland.ValidateWriteQueryResponse{}
 	response.TableID = tableID.String()
--- a/internal/tableland/impl/mesa.go
+++ b/internal/tableland/impl/mesa.go
@@ -49,7 +49,7 @@ func (t *TablelandMesa) ValidateCreateTable(
 	return tableland.ValidateCreateTableResponse{StructureHash: createStmt.GetStructureHash()}, nil
 }
 
-// RelayWriteQuery allows the user to rely on the validator wrapping the query in a chain transaction.
+// ValidateWriteQuery allows the user to validate a write query.
 func (t *TablelandMesa) ValidateWriteQuery(
 	ctx context.Context,
 	req tableland.ValidateWriteQueryRequest) (tableland.ValidateWriteQueryResponse, error) {
@@ -135,7 +135,6 @@ func (t *TablelandMesa) RunReadQuery(
 func (t *TablelandMesa) GetReceipt(
 	ctx context.Context,
 	req tableland.GetReceiptRequest) (tableland.GetReceiptResponse, error) {
-
 	if err := (&common.Hash{}).UnmarshalText([]byte(req.TxnHash)); err != nil {
 		return tableland.GetReceiptResponse{}, fmt.Errorf("invalid txn hash: %s", err)
 	}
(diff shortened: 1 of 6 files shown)


jsign:
@joewagner, I know this is draft, but just signaling that it's looking good.

joewagner:
> @joewagner, I know this is draft, but just signaling that it's looking good.

Thanks! glad to get feedback early and often.  I have a couple questions that I'll leave as inline comments, no hurry on reviewing

internal/tableland/impl/mesa.go
```diff
@@ -49,6 +49,37 @@ func (t *TablelandMesa) ValidateCreateTable(
...
+	chainID, ok := ctxChainID.(tableland.ChainID)
+	if !ok {
+		return tableland.ValidateWriteQueryResponse{}, errors.New("no chain id found in context")
+	}
+	_, chainOk := t.chainStacks[chainID]
+	if !chainOk {
+		return tableland.ValidateWriteQueryResponse{}, fmt.Errorf("chain id %d isn't supported in the validator", chainID)
+	}
+
+	mutatingStmts, err := t.parser.ValidateMutatingQuery(req.Statement, chainID)
```
joewagner, line 71: `t.parser.ValidateMutatingQuery` does not appear to check if the table exists.
I wrote an e2e test [here](https://github.com/tablelandnetwork/local-tableland/blob/joe/norelay-writes/test/test.js#L185) that will call this with a table name that doesn't exist. The call to the sdk's `write`does not fail as expected.
It seems good if we can prevent the client from paying for a statement that will fail if it isn't too difficult on our end.
@jsign would there be a way to check the tablename's existence here without too much overhead?
An alternative is we could respond from this request with the tablename in addition to the tableID, then the SDK could do a second request for a read from the table, e.g. `select * from ${tablename} limit 1;`

internal/tableland/tableland.go
```diff
@@ -58,6 +58,16 @@ type ValidateCreateTableResponse struct {
...
 }
 
+// ValidateWriteQueryRequest is a ValidateWriteQuery request.
+type ValidateWriteQueryRequest struct {
+	Statement string `json:"statement"`
+}
+
+// ValidateWriteQueryResponse is a ValidateWriteQuery response.
+type ValidateWriteQueryResponse struct {
+	TableID string `json:"table_id"`
```
joewagner, line 68: Since TableID is a big.Int it seemed natural to convert to string, but I'm not sure if there's a better approach

pkg/sqlstore/impl/system_store_instrumented.go
```diff
@@ -208,6 +209,7 @@ func (s *InstrumentedSystemStore) Begin(ctx context.Context) (*sql.Tx, error) {
 func (s *InstrumentedSystemStore) GetReceipt(
 	ctx context.Context,
 	txnHash string) (eventprocessor.Receipt, bool, error) {
+	log.Debug().Str("hash", txnHash).Msg("call GetReceipt")
```
joewagner, line 212: This is unrelated to the new rpc endpoint, but being able to see the `tableland_getReceipt` rpc call in DEBUG mode will be convenient when we implement  https://www.notion.so/textile/Handle-receipt-confirmations-in-JS-SDK-82da86b6184f4472a42d697b5ef91f9f

tableland-openapi-spec.yaml
```diff
@@ -46,6 +46,14 @@ paths:
                   params:
                     - id: "0"
                       create_statement: "CREATE TABLE myname_31337 (column_a int, column_b text);"
+              validateWriteQuery:
```
joewagner, line 49: The Local Tableland tests that are auto generated from this file all pass with these addtions, but doesn't hurt to double check if anything looks incorrect

internal/tableland/impl/mesa.go
jsign: Correct, `ValidateMutatingQuery` is a feature at the parsing level, so it's very light and fast to run.

> @jsign would there be a way to check the tablename's existence here without too much overhead?

Yes, we could change L66 to `stack , chainOk := t.chainStacks[chainID]`, and then do `_, err := stack.GetTable(ctx, tableID)`. Today, that `GetTable` would fail if that table doesn't exist but we don't have any "sentinel error" to do a concrete check if the failure is for that reason. If we add that, we could do:
```
if _, err := stack.GetTable(ctx, tableID); err == ErrTableNotFound {
   return tablleand.validateWriteQueryResponse{}, errrors.New("table doesn't exist")
}
```
The `ErrTableNotFound` is an error that doesn't exist yet. If you want, I can add that to your branch here.

Maybe might be worth adding an extra flag in the request body `OnlyParse bool` or similar, if somebody only wants to do syntax parsing? I'm not sure about this, so mostly thinking out loud. I'm not sure if anyone would like to prepare and check write-query statements before they actually create the table. If the usual use case is usually assuming the table should exist, the `OnlyParse` default value would be `false`. So this other path would be mostly optional.

internal/tableland/tableland.go
jsign: Yes, this is correct. `string` is fine since the number can be theoretically big, and in JS numbers are floats.

joewagner (commented):
@jsign I added in some more checks around table existence and left a comment on the Sentinel Errors.  Feel free to commit on this branch if you want to add or fix anything

internal/tableland/impl/mesa.go
```diff
@@ -49,6 +49,52 @@ func (t *TablelandMesa) ValidateCreateTable(
...
+
+	table, err := stack.Store.GetTable(ctx, tableID)
+
+	// if the tableID is not valid err will exist
+	if err != nil {
+		return tableland.ValidateWriteQueryResponse{}, fmt.Errorf("getting table: %s", err)
+	}
+	// if the prefix is wrong the statement is not valid
+	prefix := mutatingStmts[0].GetPrefix()
+	if table.Prefix != prefix {
```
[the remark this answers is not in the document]
jsign: @joewagner, cool, looks good. Yeah, the sentinel error is only if we want to show a "nicer" message if it's erroring in that case. We can add it later if we think it's a good idea, but what you did I believe is enough.

jsign (approved):

jsign:
@joewagner, BTW, congrats on the PR. Happy to see more help from you here!

joewagner:
> @joewagner, BTW, congrats on the PR. Happy to see more help from you here!

Thanks!  Thanks especially for all the guidance and help.  It's fun to be able to work on this, hopefully more to come

PR #177, opened 2022-07-12
Create cmd/walletcreator & use gofumpt in lints and files

jsign:
While writing validator documentation, I realized it would be nice to provide a simple and transparent tool to allow people to create wallet addresses. They can still use external tools like Metamask or others, but that sounds complicated and hard to audit compared to run a <10 LOC CLI tool that you can compile yourself.

I also:
- Make `gofumpt` part of our linting checks.
- Fix all files to comply to `gofumpt`
- Update other pending deps.


.golangci.yml
```diff
@@ -15,6 +15,7 @@ linters:
     - godot
     - lll
     - sqlclosecheck
+    - gofumpt
```
jsign, line 18: `make lint` (i.e: GH Action linter) will check for `gofumpt` formatting.

asutula (approved):

PR #180, opened 2022-07-13
otel: automatically instrument database/sql connections

jsign:
In this PR we add a library to wrap every opened connection to a `database/sql` driver to have a set of extra metrics at this layer. More about this at: https://github.com/XSAM/otelsql

I also updated the dashboard with:
- I collapsed in the same section the `HTTP requests` and `Public APIs` since they both are about the same things.
- I reordered sections a bit to put into the top more relevant ones.
- I renamed the section "SQL Store calls" with "SQL Database" and added two new tiles in the second row related to the new metrics of this wrapper (see screenshot below). You can see the live staging dashboard too.
 
image
(Not entirely interesting since it's staging)


brunocalza (approved):
👍

PR #183, opened 2022-07-15
grafana/dashboard: add #goroutines, liveobjects and gc/s charts & general improvements

jsign:
Now that we see more read volume, I'm adding some extra charts in the dashboard reg:
- #goroutines
- #live objects
- GC/s
The first two to spot any trends in leaked goroutines or mem objects.

 Taking advantage and fixing other things:
- Better distinction between req/s and latency of REST paths vs JSON-RPC calls.
- We have a counter of any API response size, so we now we show ~"API output bytes / s"
- Fixed many legends

Also, I removed adding the `controller` in the metrics dimension. We aren't using this dimension, and it has the risk of exploding dimensions in Prometheus. If we come up with a chart that really uses the dimension, we can add it again.

### DIFF
--- a/internal/system/impl/sqlstore_instrumented.go
+++ b/internal/system/impl/sqlstore_instrumented.go
@@ -69,10 +69,8 @@ func (s *InstrumentedSystemSQLStoreService) GetTablesByController(ctx context.Co
 	latency := time.Since(start).Milliseconds()
 	chainID, _ := ctx.Value(middlewares.ContextKeyChainID).(tableland.ChainID)
 
-	// NOTE: we may face a risk of high-cardilatity in the future. This should be revised.
 	attributes := []attribute.KeyValue{
 		{Key: "method", Value: attribute.StringValue("GetTablesByController")},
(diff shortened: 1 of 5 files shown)


internal/system/impl/sqlstore_instrumented.go
```diff
@@ -69,10 +69,8 @@ func (s *InstrumentedSystemSQLStoreService) GetTablesByController(ctx context.Co
 	latency := time.Since(start).Milliseconds()
 	chainID, _ := ctx.Value(middlewares.ContextKeyChainID).(tableland.ChainID)
 
-	// NOTE: we may face a risk of high-cardilatity in the future. This should be revised.
 	attributes := []attribute.KeyValue{
 		{Key: "method", Value: attribute.StringValue("GetTablesByController")},
-		{Key: "controller", Value: attribute.StringValue(controller)},
```
brunocalza, line 75: So nice you remembered about this 👍

brunocalza (approved):
💯

PR #185, opened 2022-07-15
Adds SetController call to toolkit

brunocalza:
Also, removes the param `caller` from open api spec.

### DIFF
--- a/tableland-openapi-spec.yaml
+++ b/tableland-openapi-spec.yaml
@@ -81,8 +81,7 @@ paths:
                   method: tableland_setController
                   id: 1
                   params:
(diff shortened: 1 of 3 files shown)


tableland-openapi-spec.yaml
```diff
@@ -81,8 +81,7 @@ paths:
                   method: tableland_setController
                   id: 1
                   params:
-                    - caller: "0xf39fd6e51aad88f6f4ce6ab8827279cfffb92266"
```
brunocalza, line 84: Femoving from openapi spec
cc @dtbuchholz

jsign (approved):

PR #186, opened 2022-07-15
router: use uninstrumented mesa for UserController

jsign:
The `UserController` was taking as a dependency an instrumented `mesa`, which generates a correlation between the `/query` REST API and the `RunReadQuery` JSON-RPC. This is because `/query` calls the RPC method internally.

This has an undesired side-effect of correlating both RPCs in the metrics:
image

Calls that come from `/query` should only be reflected in `/query`, not indirectly in `RunReadQuery`.

The fix here is to avoid using the instrumented `mesa`, thus any call from `UserController` to the `mesa` service wouldn't _pollute_ the middleware metrics.

### DIFF
--- a/internal/router/router.go
+++ b/internal/router/router.go
@@ -33,12 +33,16 @@ func ConfiguredRouter(
 		log.Fatal().Err(err).Msg("creating instrumented user store")
 	}
 
-	svc := getTablelandService(parser, instrUserStore, chainStacks)
+	mesaService := impl.NewTablelandMesa(parser, instrUserStore, chainStacks)
+	userController := controllers.NewUserController(mesaService)
+	mesaService, err = impl.NewInstrumentedTablelandMesa(mesaService)
+	if err != nil {
+		log.Fatal().Err(err).Msg("instrumenting mesa")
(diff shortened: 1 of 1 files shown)


internal/router/router.go
```diff
@@ -90,20 +94,6 @@ func ConfiguredRouter(
 	return router
 }
 
-func getTablelandService(
```
jsign, line 93: No need for this helper method.

internal/router/router.go
```diff
@@ -33,12 +33,16 @@ func ConfiguredRouter(
 		log.Fatal().Err(err).Msg("creating instrumented user store")
 	}
 
-	svc := getTablelandService(parser, instrUserStore, chainStacks)
+	mesaService := impl.NewTablelandMesa(parser, instrUserStore, chainStacks)
+	userController := controllers.NewUserController(mesaService)
```
jsign, line 37: Use the uninstrumented first to create `userController`, and then instrument it and build `mesaService`.

brunocalza (approved):
Looks good!

Issue #187, opened 2022-07-17
REST endpoint doesn't return any tables for a controller's address


dtbuchholz:
### User Story
When I call the `/chain/{chainId}/tables/controller/{address}` endpoint, I expect a response of an array of objects where each object represents a table controlled at the path param `address`. However, when calling this endpoint using a smart contract address that implements `TablelandController` and has been set as the controller for a table, the results are empty.

### Steps to Reproduce
_The following was performed on Goerli (chainId: `5`)_
1. Create a table
2. Deploy a smart contract that implements `TablelandController` ([allow all example](https://github.com/tablelandnetwork/evm-tableland/blob/main/contracts/test/TestAllowAllTablelandController.sol) from the `evm-tableland` repo)
3. Call `setController` on the `TablelandTables` deployed smart contract with the address from (2) and the table from (1)
4. Insert data into the table with a random wallet, which _should_ be able to due to the allow all contract (this works as expected)
5. Call the REST API at endpoint `/chain/{chainId}/tables/controller/{address}` -- here, the response is empty, i.e., `[]`
6. Validate the controller address is the controller by calling `getController` on the registry smart contract with the tableId from (1), which should match the address from (2)

Here's an example that does the steps above:
1. See table `rest_api_5_85` (created at tx: [`0xf8da1143d65fde76c3a175685810146ea32d6695af76cf1d77758e879abaa321`](https://goerli.etherscan.io/tx/0xf8da1143d65fde76c3a175685810146ea32d6695af76cf1d77758e879abaa321))
2. See the deployed controller contract at address [`0x627A289FDfFA63b662FC52FE8b30b479B8AED2a8`](https://goerli.etherscan.io/address/0x627A289FDfFA63b662FC52FE8b30b479B8AED2a8), which simply implements the provided allow all contract in the `evm-tableland` repo (link above in step 2)
3. Calling `setController` with the address above on the [Tableland registry contract for Goerli](https://goerli.etherscan.io/address/0xda8ea22d092307874f30a1f277d1388dca0ba97a#readProxyContract) -- tx hash: [`0x83f1f3204cc8df3b4e27789579bbf485af566cbaa1a26bc2cf8f2c1556989c69`](https://goerli.etherscan.io/tx/0x83f1f3204cc8df3b4e27789579bbf485af566cbaa1a26bc2cf8f2c1556989c69)
[...]

PR #190, opened 2022-07-19
Fixes transfer event execution

brunocalza:
# Summary

The execution of a transfer event was not changing the table's owner in the registry table.

# Context

[Two tables were transferred on Goerli](https://goerli.etherscan.io/address/0x0C53611b433B486ec940334653F7756D75147dEA#tokentxnsErc721) (57 and 59) to address `0x0C53611b433B486ec940334653F7756D75147dEA`, but when executing the call `https://testnet.tableland.network/chain/5/tables/controller/0x0C53611b433B486ec940334653F7756D75147dEA` you could not see those tables listed.

The undesired behavior happened because of a missing logic when executing the transfer event logic. 

## More info

- I already synced from scratch locally (pointing to `testnet` contract) and the logic worked. 
- Deployed to staging
- Even after this PR is merged and deployed to `testnet`, we'll need to sync all events from scratch to pick up the change (that will probably be done in a future time)

cc @sanderpick

### DIFF
--- a/pkg/eventprocessor/impl/eventprocessor.go
+++ b/pkg/eventprocessor/impl/eventprocessor.go
@@ -376,9 +376,9 @@ func (ep *EventProcessor) executeCreateTableEvent(
 	tableID := tableland.TableID(*e.TableId)
 
 	if err := b.InsertTable(ctx, tableID, e.Owner.Hex(), createStmt); err != nil {
-		var pgErr *txn.ErrQueryExecution
-		if errors.As(err, &pgErr) {
-			err := fmt.Sprintf("table creation execution failed (code: %s, msg: %s)", pgErr.Code, pgErr.Msg)
+		var dbErr *txn.ErrQueryExecution
+		if errors.As(err, &dbErr) {
+			err := fmt.Sprintf("table creation execution failed (code: %s, msg: %s)", dbErr.Code, dbErr.Msg)
 			receipt.Error = &err
 			return receipt, nil
 		}
@@ -418,9 +418,9 @@ func (ep *EventProcessor) executeRunSQLEvent(
 		return receipt, nil
 	}
 	if err := b.ExecWriteQueries(ctx, e.Caller, mutatingStmts, e.IsOwner, &policy{e.Policy}); err != nil {
-		var pgErr *txn.ErrQueryExecution
-		if errors.As(err, &pgErr) {
-			err := fmt.Sprintf("db query execution failed (code: %s, msg: %s)", pgErr.Code, pgErr.Msg)
+		var dbErr *txn.ErrQueryExecution
+		if errors.As(err, &dbErr) {
+			err := fmt.Sprintf("db query execution failed (code: %s, msg: %s)", dbErr.Code, dbErr.Msg)
 			receipt.Error = &err
 			return receipt, nil
 		}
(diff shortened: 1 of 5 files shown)


pkg/eventprocessor/impl/eventprocessor.go
```diff
@@ -487,22 +487,32 @@ func (ep *EventProcessor) executeTransferEvent(
 		receipt.Error = &tableIDIsEmpty
 		return receipt, nil
 	}
+
 	tableID := tableland.TableID(*e.TableId)
+	if err := b.ChangeTableOwner(ctx, tableID, e.To); err != nil {
```
brunocalza, line 492: This was the missing logic responsible for the undesired behavior

pkg/txn/impl/processor_test.go
```diff
@@ -804,6 +806,48 @@ func TestWithCheck(t *testing.T) {
...
 }
 
+func TestChangeTableOwner(t *testing.T) {
+	t.Parallel()
+	ctx := context.Background()
+
+	tableID := tableland.TableID(*big.NewInt(100))
+	txnp, _, db := newTxnProcessorWithTable(t, 0)
+
+	require.Equal(t, 1,
```
brunocalza, line 816: Checks that there's an entry for old owner

pkg/txn/impl/processor_test.go
```diff
@@ -804,6 +806,48 @@ func TestWithCheck(t *testing.T) {
...
+	require.NoError(t, err)
+
+	// change table's owner
+	err = b.ChangeTableOwner(ctx, tableID, controller)
+	require.NoError(t, err)
+	require.NoError(t, b.Commit())
+	require.NoError(t, b.Close())
+	require.NoError(t, txnp.Close(ctx))
+
+	require.Equal(t, 1,
```
brunocalza, line 838: Checks that there's an entry for new owner

pkg/eventprocessor/impl/eventprocessor.go
```diff
@@ -376,9 +376,9 @@ func (ep *EventProcessor) executeCreateTableEvent(
 	tableID := tableland.TableID(*e.TableId)
 
 	if err := b.InsertTable(ctx, tableID, e.Owner.Hex(), createStmt); err != nil {
-		var pgErr *txn.ErrQueryExecution
```
brunocalza, line 379: Just changing `pg` to `db` in a bunch of places.

pkg/txn/impl/processor.go
```diff
@@ -420,6 +420,38 @@ func (b *batch) TxnReceiptExists(ctx context.Context, txnHash common.Hash) (bool
 	return exists, nil
 }
 
+// ChangeTableOwner changes the owner of the table in the registry table.
+func (b *batch) ChangeTableOwner(
```
brunocalza, line 424: Opted for the `Owner` name instead of `controller` (name of the column). Looks like the name `controller` is now more related to the Controller smart contract.

jsign: Q: Worth renaming the `registry` table column?

jsign (approved):
LGTM!

pkg/txn/impl/processor.go
brunocalza: I thought about it. I think it is but wouldn't do in this PR. There's some discussion [here](https://github.com/tablelandnetwork/go-tableland/issues/187) about how `controller` is misleading. Maybe we can take advantage of what will come up there to do this change.

jsign: :+1:

Issue #170, opened 2022-07-09
external_url for tables endpoint is inconsistent


sanderpick:
eg,

https://testnet.tableland.network/chain/69/tables/6: `external_url` doesn't include `/chain/{chainID}/`

on localhost,

`external_url` is reported as `https://testnet.tableland.network/{tableID}`:
- host should be localhost
- `/chain/{chainID}/tables/` is missing

PR #199, opened 2022-07-21
Fixes external url

brunocalza:
Solves issue #170

### DIFF
--- a/docker/deployed/staging/api/config.json
+++ b/docker/deployed/staging/api/config.json
@@ -7,7 +7,7 @@
     "TLSKey": "${VALIDATOR_TLS_KEY}"
   },
   "Gateway": {
-    "ExternalURIPrefix": "https://staging.tableland.network/tables"
+    "ExternalURIPrefix": "https://staging.tableland.network"
   },
   "DB": {
     "Port": "5432"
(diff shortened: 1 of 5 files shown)


docker/local/api/config.json
```diff
@@ -3,6 +3,9 @@
     "Human": true,
     "Debug": true
   },
+  "Gateway": {
```
brunocalza, line 6: Adds this to `local`. Without this config, the `default` points to `https://testnet.tableland.network` which is weird

docker/deployed/staging/api/config.json
```diff
@@ -7,7 +7,7 @@
     "TLSKey": "${VALIDATOR_TLS_KEY}"
   },
   "Gateway": {
-    "ExternalURIPrefix": "https://staging.tableland.network/tables"
```
brunocalza, line 10: `/tables` is removed because it's part of the [Sprintf](https://github.com/tablelandnetwork/go-tableland/pull/199/files?diff=unified&w=0#diff-495917a0bcbd30f16dabee67340414731bcf65805956d9209a21da47227885e6R66) now.

jsign (approved):
LGTM

PR #200, opened 2022-07-21
tls: tune configuration for CF

jsign:
I got a message from CF that we might be accepting TLS ciphers that aren't aligned with what they expect.

Now set ciphers to match [their official documentation](https://developers.cloudflare.com/ssl/origin-configuration/cipher-suites/) to see if this impacts them.

After we figure out if it does (or not), we can see later to revert or make this an optional flag. For now, take this as giving it a try to fix the 520 from CF.


brunocalza (approved):
Nice. Crossing fingers!

PR #208, opened 2022-07-29
Adds GET /schema and /tables/structure endpoints [stagingdeploy]

brunocalza:
This PR implements 2 new REST endpoints:
1. `GET /schema{table_name}`
2. `GET /chain/id/tables/structure/{hash}`

The first endpoint solves de **schema discovery** problem: **given a table, what's the schema**? The second endpoint solves the inverse problem: **given a schema, what are the tables that share that schema**?

The implementations are very straightforward, just making use of the same code architecture used for other endpoints.
- In 1, we get the schema by executing `SELECT sql FROM sqlite_master WHERE name=?1;` and sending the response of that query to our parser
- In 2, is an analogous implementation of `GET /chain/id/tables/controller/{controller}`

---
A side note about this PR: the implementations of two trivial endpoints took too much time and touched too many files. Usually, a signal that something is off in the architecture or too complicated for a simple feature. One part that is clearly strange is the db layers. Since we don't use `sqlc` more, I don't think there's a need to keep using `sqlc` structure. I don't think there's something worth doing right now, just mentioning to keep it in mind.


internal/router/controllers/system.go
```diff
@@ -83,21 +81,117 @@ func (c *SystemController) GetTablesByController(rw http.ResponseWriter, r *http
...
+			Structure:  t.Structure,
+		}
+	}
+
+	rw.WriteHeader(http.StatusOK)
+	_ = json.NewEncoder(rw).Encode(retTables)
+}
+
+// GetSchemaByTableName handles the GET /schema/{table_name} call.
+func (c *SystemController) GetSchemaByTableName(rw http.ResponseWriter, r *http.Request) {
```
brunocalza, line 141: The second endpoint: `GET /schema/{table_name}`

internal/router/controllers/system.go
```diff
@@ -83,21 +81,117 @@ func (c *SystemController) GetTablesByController(rw http.ResponseWriter, r *http
...
+			Error().
+			Err(err).
+			Str("name", name).
+			Msg("failed to fetch tables")
+
+		_ = json.NewEncoder(rw).Encode(errors.ServiceError{Message: "Failed to get schema from table"})
+		return
+	}
+
+	if len(schema.Columns) == 0 {
```
brunocalza, line 161: This is a check for empty schema. Unfortunately, `if (sqlstore.Schema{}) == schema` is not an option because the fields of `sqlstore.Schema` are not comparable.

internal/router/controllers/system_test.go
```diff
@@ -14,27 +14,115 @@ import (
...
-
-	//nolint
-	expJSON := `{
-		"name":"name-1",
-		"external_url":"https://tableland.network/tables/100",
-		"image":"https://bafkreifhuhrjhzbj4onqgbrmhpysk2mop2jimvdvfut6taiyzt2yqzt43a.ipfs.dweb.link",
-		"attributes":[{"display_type":"date","trait_type":"created","value":1546360800}]
-	}`
-	require.JSONEq(t, expJSON, rr.Body.String())
+	t.Run("get table metadata", func(t *testing.T) {
```
brunocalza, line 20: As a reminder, the tests here use a mocked implementation. So what is being tested is whether the request and response are as expected.

internal/tableland/tableland.go
```diff
@@ -134,3 +137,40 @@ func NewTableIDFromInt64(intID int64) (TableID, error) {
...
+	chainID ChainID
+}
+
+// ChainID returns table's chain id.
+func (t Table) ChainID() ChainID {
+	return t.chainID
+}
+
+// NewTableFromName creates a Table from its name.
+func NewTableFromName(name string) (Table, error) {
```
brunocalza, line 154: This method is used to validate a table name and get the chain id.

tableland-openapi-spec.yaml
```diff
@@ -286,6 +320,57 @@ paths:
...
+  /schema/{tableName}:
+    get:
+      summary: Get the schema of a table by its full name
+      parameters:
+        - in: path
+          name: tableName
+          schema:
+            type: string
+          required: true
+          description: The full table of the name
```
jsign, line 332: Nit: maybe wanted to write "Full name of the table", or "Full table name" :smile:

jsign (approved):
Nice! LGTM.

I left some comments for your consideration!

pkg/sqlstore/impl/system/store.go
```diff
@@ -201,6 +202,72 @@ func (s *SystemStore) ReplacePendingTxByHash(ctx context.Context, oldHash common
...
+// GetSchemaByTableName get the schema of a table by its name.
+func (s *SystemStore) GetSchemaByTableName(ctx context.Context, name string) (sqlstore.TableSchema, error) {
+	createStmt, err := s.db.queries().GetSchemaByTableName(ctx, db.GetSchemaByTableNameParams{
+		TableName: name,
+	})
+	if err != nil {
+		return sqlstore.TableSchema{}, fmt.Errorf("failed to get the table: %s", err)
+	}
+
+	createStmt = strings.ReplaceAll(createStmt, "STRICT", "")
```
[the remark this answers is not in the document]
brunocalza: Good callout

pkg/sqlstore/impl/system/store.go
```diff
@@ -201,6 +202,72 @@ func (s *SystemStore) ReplacePendingTxByHash(ctx context.Context, oldHash common
...
+// GetSchemaByTableName get the schema of a table by its name.
+func (s *SystemStore) GetSchemaByTableName(ctx context.Context, name string) (sqlstore.TableSchema, error) {
+	createStmt, err := s.db.queries().GetSchemaByTableName(ctx, db.GetSchemaByTableNameParams{
+		TableName: name,
+	})
+	if err != nil {
+		return sqlstore.TableSchema{}, fmt.Errorf("failed to get the table: %s", err)
+	}
+
+	index := strings.LastIndex(createStmt, "STRICT")
```
brunocalza, line 235: @jsign lmk if this is a safer approach

jsign: Yep :+1:

PR #213, opened 2022-08-03
Multi-event transaction support & other improvements

jsign:
This PR implements support for multi-events in EVM transactions. 

The TL;DR is that EVM transactions can have multiple events, which will be executed in an all-or-nothing style allowing more powerful composability and atomicity for users.

Related changes to this feature are:

- Make EventFeed do the event grouping per EVM txn.
- The EventProcessor now has better high-level abstractions to do this processing:
    - `Executor` allows executing blocks for a chain. It creates `BlockScopes` to give the `EventProcessor` a scratchpad to execute txn with their events. The `Executor` abstraction can be the basis for future multi-versioned execution pipelines that will be needed.
    - `BlockScope` is a scope to do all the work in a particular block. (Previously called `Batch`).
    - `TxnScope` is a scope nested in a `BlockScope` to execute a list of events in an all-or-nothing style.
    - All previous APIs for executing events and receipts were mixed in the `Batch` abstraction. Now they’re separated into `TxnScope`, `BlockScope`, and `Executor`.
    - All these new “scopes” have a single txn to execute operations, so there’s no ambiguity about receiving `txn` via parameters or any possible confusion about using a parent transaction.
- I had to do a big refactor of tests to adjust to the new APIs and, of course, further tests to see if the new feature worked correctly.
- I split logic into separate files guided by event types. This allows finding faster the logic related to processing a particular type of event and organizing future logic.
- The `TxnScope` has its own annotated logger with the transaction hash, which provides all the logs related to executing a particular transaction (or block) more cleanly.
- The API for getting the EVM transaction receipt has a new field `error_event_idx`, indicating the event index that has produced a failure. Only applicable if the transaction errored, or `-1` otherwise.

Apart from this main change, I took advantage and made many other improvements:

- The event processor, the channel with the event feed, is now buffered. This is safe to do since the processing is still serial. The buffering allows delays or variability in the event feed talking, with Alchemy not making the event processor idle (and the reverse). I did a full cold sync, and locally it took 4 min to do a full sync (instead of the previously reported ~10min).
- I standardized log attributes to snake case since we have many places where we use names like `chainID`, `chainid` or `chainId`. So when looking at logs for all logs related to a chain-id, you can miss logs depending on what name you choose. Now all should be snake-case (e.g. “chain_id”, “block_number”, etc.).
- All `database/sql` connections have `MaxIdleConnections(0)` now to confirm this allows the WAL checkpointing to happen. I also renamed `pool` names to `db` since now there’s no real pooling being done.

I’ll add more details as PR review comments.


pkg/eventprocessor/impl/executor/impl/executor_test.go
```diff
@@ -0,0 +1,265 @@
+package impl
```
jsign, line 1: You'll see that test files now are organized in different files. We had many of them bundled in the same file and was very hard to track "what was where".

Many of them are the original ones with everything fix to use the new pipeline. 
I'll make some general comments on relevant changes in the wild.

pkg/eventprocessor/impl/executor/impl/executor_test.go
```diff
@@ -0,0 +1,265 @@
...
+	ok, err = bs.TxnReceiptExists(ctx, common.HexToHash(txnHash))
+	require.NoError(t, err)
+	require.True(t, ok)
+	require.NoError(t, bs.Commit())
+	require.NoError(t, bs.Close())
+
+	require.NoError(t, ex.Close(ctx))
+}
+
+func TestMultiEventTxnBlock(t *testing.T) {
```
jsign, line 64: This new test thoroughly tests a complex scenario with multiple txns in the same block, each with multiple transactions.

Note that the way the code is architected now, we can create e2e tests with this although the SC doesn't have any new method to create this situation from SC calls. 

Let's take a look here...

pkg/eventprocessor/impl/executor/impl/executor_test.go
```diff
@@ -0,0 +1,265 @@
...
+
+	// The test setup is executing a block that has the following transactions:
+	// - Txn 1, successful with two events: CREATE a table and INSERT a row.
+	// - Txn 2, failed with two events: CREATE a table and INSERT a row. INSERT has invalid SQL.
+	// - Txn 3, successful with one event: INSERT a row in the table created in Txn 1.
+	//
+	// Checks after block execution:
+	// 1) Txn 1 side-effects are seen in the db.
+	// 2) Txn 2 side-effects aren't seen in the db. In particular, the table created should have been rollbacked.
+	// 3) Txn 3 side-effects are seen in the db. Txn 2 failing should be isolated and have no effect in Txn 3 execution.
```
jsign, line 75: This gives a complete explanation of what is the idea of the tests.
LMK if something isn't clear.

TL;DR: We'll execute 3 transactions in a block. Each of them with two events. Transactions 1 and 3 will succeed. Transaction 2 will fail since the **second** event fails and the first succeeds.

This produces an interesting situation where:
- Txn1 and Txn2 should affect the state of the DB.
- Txn2 should be rollbacked. And extra check that the first event also was rollbacked despite its success.

pkg/eventprocessor/impl/executor/impl/executor_test.go
```diff
@@ -0,0 +1,265 @@
...
+	ctx := context.Background()
+	ex, dbURI := newExecutor(t, 0)
+
+	bs, err := ex.NewBlockScope(ctx, 0)
+	require.NoError(t, err)
+
+	// Txn 1
```
jsign, line 82: I separated each Txn in a `{  ... } ` scope. You'll see the usual new mechanics of running tests reg creating txns and events.

pkg/eventprocessor/impl/executor/impl/executor_test.go
```diff
@@ -0,0 +1,265 @@
...
+		})
+		require.NoError(t, err)
+		// This Txn should fail.
+		require.NotNil(t, res.Error)
+		require.Equal(t, 1, *res.ErrorEventIdx)
+	}
+	// Txn 3
+	{
+		// We can leverage helper for single event txn.
+		assertExecTxnWithRunSQLEvents(t, bs, []string{"insert into bar_1337_100 values ('txn 3')"})
```
jsign, line 141: I've created some `assertExecTxnWithXXXX(...)` helpers for each event type. These helpers simplify executing transactions with single events to avoid boilerplate in 80% of tests.

In more complex scenarios, such as the two Txn1 and Txn2 scopes above, you can use the full API as used there.

pkg/eventprocessor/impl/executor/impl/txnscope.go
```diff
@@ -0,0 +1,109 @@
...
+
+type txnScope struct {
+	log zerolog.Logger
+
+	parser    parsing.SQLValidator
+	acl       tableland.ACL
+	scopeVars scopeVars
+
+	txn *sql.Tx
+}
```
jsign, line 41: We're in `txnscope.go` here. This has the same idea of `blockscope.go`.

pkg/eventprocessor/impl/executor/impl/txnscope.go
```diff
@@ -0,0 +1,109 @@
...
+
+type eventExecutionResult struct {
+	TableID *tableland.TableID
+	Error   *string
+}
+
+func (ts *txnScope) executeTxnEvents(
```
jsign, line 48: Now the event type detection and routing to execution logic happens at this layer, since it's a responsibility at the EVM transaction scope.

pkg/eventprocessor/impl/executor/impl/txnscope.go
```diff
@@ -0,0 +1,109 @@
...
+	evmTxn eventfeed.TxnEvents,
+) (executor.TxnExecutionResult, error) {
+	var res eventExecutionResult
+	var err error
+
+	for idx, event := range evmTxn.Events {
+		switch event := event.(type) {
+		case *ethereum.ContractRunSQL:
+			ts.log.Debug().Str("statement", event.Statement).Msgf("executing run-sql event")
+			res, err = ts.executeRunSQLEvent(ctx, event)
```
jsign, line 59: Note that all these `ts.executeXXXXEvent(...)` have straightforward arguments. You can check in the old implementation we had 4 or 5 arguments and it was pretty complicated.

pkg/eventprocessor/impl/executor/impl/txnscope.go
```diff
@@ -0,0 +1,109 @@
...
+
+		// If the current event fail, we stop processing further events in this transaction and already
+		// return the failed receipt. This receipt contains the index of this failed event.
+		if res.Error != nil {
+			return executor.TxnExecutionResult{
+				TableID:       res.TableID,
+				Error:         res.Error,
+				ErrorEventIdx: &idx,
+			}, nil
+		}
```
jsign, line 105: Depending on the resutlt of the execution, we create the `TxnExecutionResult` that will be used by the `EventProcessor` to build the `TransactionReceipt` (as explained some comments ago).

pkg/eventprocessor/impl/executor/impl/txnscope.go
```diff
@@ -0,0 +1,109 @@
...
+			return executor.TxnExecutionResult{}, fmt.Errorf("unknown event type %t", event)
+		}
+
+		// If the current event fail, we stop processing further events in this transaction and already
+		// return the failed receipt. This receipt contains the index of this failed event.
+		if res.Error != nil {
+			return executor.TxnExecutionResult{
+				TableID:       res.TableID,
+				Error:         res.Error,
+				ErrorEventIdx: &idx,
```
jsign, line 103: This is a new field.

pkg/eventprocessor/impl/executor/impl/txnscope_createtable.go
```diff
@@ -0,0 +1,81 @@
+package impl
```
jsign, line 1: Note how I separate the business logic of each event type in `txnscope_xxxxx.go`.
This allows finding fast all the logic corresponding to particular event execution. 

I did the same with tests; they live in `txnscope_xxxx_test.go`, so we can also find the tests related to an event type quite fast.

pkg/eventprocessor/impl/executor/impl/txnscope_createtable.go
```diff
@@ -0,0 +1,81 @@
...
+
+	"github.com/textileio/go-tableland/internal/tableland"
+	"github.com/textileio/go-tableland/pkg/parsing"
+	"github.com/textileio/go-tableland/pkg/tables/impl/ethereum"
+)
+
+func (ts *txnScope) executeCreateTableEvent(
+	ctx context.Context,
+	e *ethereum.ContractCreateTable,
+) (eventExecutionResult, error) {
```
jsign, line 16: You'll note that each method that has the logic to process an event type returns a `eventExecutionResult` (see [here](https://github.com/tablelandnetwork/go-tableland/blob/750cb2793dfaaefe42d119d9c7aa1a7c04ed5631/pkg/eventprocessor/impl/executor/impl/txnscope.go#L43)). 

This is even a more constrained struct compared to `TxnExecutionResult`. It only has the `TableID` and the potential `Error`. This is necessary since we can't exactly create a `TxnExecutionResult` in this layer, since we don't know which is the index of the event in the transaction. We're only interested in executing an event here. No need to pollute with more arguments that are only useful to build some high-level struct.

pkg/eventprocessor/impl/executor/impl/txnscope_createtable_test.go
```diff
@@ -0,0 +1,67 @@
...
+	t.Run("success", func(t *testing.T) {
+		t.Parallel()
+		ctx := context.Background()
+
+		ex, dbURI := newExecutor(t, 0)
+
+		bs, err := ex.NewBlockScope(ctx, 0)
+		require.NoError(t, err)
+
+		assertExecTxnWithCreateTable(t, bs, 100, "0xb451cee4A42A652Fe77d373BAe66D42fd6B8D8FF", "create table bar_1337 (zar text)") //nolint
```
jsign, line 30: Another example of a helper `assertExecTxnWithXXXX(...)`. 
It makes it easy to execute an XXXX event in the pipeline, and assert all ran correctly. This allows for having shorter test logic in most places.

pkg/eventprocessor/impl/executor/impl/txnscope_runsql.go
```diff
@@ -0,0 +1,524 @@
+package impl
```
jsign, line 1: Notice how separating logic per event type in files shows the complexity of the event type execution implementation.

pkg/eventprocessor/impl/executor/impl/txnscope_runsql_test.go
```diff
@@ -0,0 +1,555 @@
+package impl
```
jsign, line 1: I've reorganized a bit all these tests:
- Different names
- I think some splits of subtests on separate tests.

Mostly because now we can have "OneEventPerTxn" or multiple (which wasn't true before).

Also, you should expect tests to feel a bit cleaner since:
- We don't do `ValidateCreateTable(..., query)` anymore
- We don't depend on internal results of the previous `Txnprocessor`.

brunocalza (changes requested):
Nice work! The code architecture is very clean, and the abstractions represent much better the nature of the problem.

I have a feeling some changes must be done at the JSON response of the receipt. Check my comment [here](https://github.com/tablelandnetwork/go-tableland/pull/213/files#r936966362). 

Other than that, everything is great.

pkg/sqlstore/impl/system/migrations/002_receipterroridx.up.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE system_txn_receipts ADD error_event_idx INTEGER DEFAULT 0;
```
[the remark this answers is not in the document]
jsign: Uhm, this default is 100% wrong. The thing here was that at this point in time, I was doubting if making it nullable or giving it a 0 value and not making it null.

The default's main point was to migrate _existing data_. In that case, those existing entries should be `null`, not 0.

I want to keep the fact that the column can be null since I want to optimize for our future API evolution that doesn't have this weird-ish -1 interpretation.

So considering that, I'll:
1) Keep the `INTEGER` that can be null.
2) Remove the `DEFAULT 0`
3) Include an extra SQL query that would do something like `UPDATE ... SET error_event_idx=0 WHERE error is not null`.

The extra point is the last one. LMK if makes sense to you to double check.

internal/tableland/tableland.go
```diff
@@ -41,8 +41,10 @@ type TxnReceipt struct {
 	ChainID     ChainID `json:"chain_id"`
 	TxnHash     string  `json:"txn_hash"`
 	BlockNumber int64   `json:"block_number"`
-	Error       *string `json:"error,omitempty"`
-	TableID     *string `json:"table_id,omitempty"`
+
+	TableID       *string `json:"table_id,omitempty"`
+	Error         *string `json:"error,omitempty"`
+	ErrorEventIdx int     `json:"error_event_idx,omitempty"`
```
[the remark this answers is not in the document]
jsign: Definitely agree on removing `omitempty` for `ErrorEventIdx`.

For `Error`, agree on that too but I wonder if this may affect some behavior in the SDKs if they already have an interpretation between "not there" and "there but empty". I'll ping @joewagner . If all is fine, I'd also remove the `omitempty` in this field too.

pkg/sqlstore/impl/system/migrations/002_receipterroridx.up.sql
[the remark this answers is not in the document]
brunocalza: Totally forgot about the migration of old rows. Now I get your intention. But yeah. the 3 steps make sense!

pkg/eventprocessor/impl/executor/impl/blockscope.go
```diff
@@ -0,0 +1,187 @@
...
+		log: logger.With().
+			Str("component", "txnscope").
+			Int64("chain_id", int64(bs.scopeVars.ChainID)).
+			Str("txn_hash", evmTxn.TxnHash.String()).
+			Logger(),
+
+		txn: bs.txn,
+	}
+	res, err := ts.executeTxnEvents(ctx, evmTxn)
+	if err != nil || res.Error != nil {
```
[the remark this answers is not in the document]
brunocalza: Oh, I thought it was returning in the first if in L89 (L91 tricked me). It is just rolling back. And then, if it is an infra error it returns the error. Makes sense. Thanks!

internal/tableland/tableland.go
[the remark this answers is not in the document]
jsign: I talked with @joewagner  and have his :+1:. So, I'll also remove the `omitempty`  from `Error`.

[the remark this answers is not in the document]
joewagner: This looks like it works with the current SDK, so we should be all good having the "there but empty" error

brunocalza (approved):
Awesome work 👏👏👏

PR #217, opened 2022-08-08
Adds SVG renderer to table metadata

brunocalza:
Adds a config for the table metadata renderer: `https://render.tableland.xyz`

cc @awmuncy

### DIFF
--- a/docker/deployed/staging/api/config.json
+++ b/docker/deployed/staging/api/config.json
@@ -8,7 +8,7 @@
   },
   "Gateway": {
     "ExternalURIPrefix": "https://staging.tableland.network",
-    "MetadataRendererURI": "https://render.tableland.xyz"
+    "MetadataRendererURI": ""
   },
   "DB": {
     "Port": "5432"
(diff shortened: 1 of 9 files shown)


docker/deployed/staging/api/config.json
```diff
@@ -7,7 +7,8 @@
     "TLSKey": "${VALIDATOR_TLS_KEY}"
   },
   "Gateway": {
-    "ExternalURIPrefix": "https://staging.tableland.network"
+    "ExternalURIPrefix": "https://staging.tableland.network",
+    "MetadataRendererURI": "https://render.tableland.xyz"
```
jsign, line 11: Shouldn't this URL be empty for staging? Because I think it would point out to testnet tables. 

If empty, prob we need to adjust the `fmt.Sprintf(...)` below to cover that case. Or see if we can make another deployment of  `https://render.tableland.xyz` for the staging env.

jsign (approved):
LGTM!

PR #219, opened 2022-08-11
Adds SVG rendering when metadata call can't find the table

brunocalza:
This PR updates the response for getting metadata in the `GET /chain/{chain_id}/tables/{table_id}` call.

Instead of returning an error, such as `"Failed to fetch metadata"` we return a response that complies to that NFT metadata standard and also an SVG to represent that visually. 

Examples: 
- Invalid chain: https://staging.tableland.network/chain/45/tables/30
- Table not found: https://staging.tableland.network/chain/69/tables/30

cc @andrewxhill

### DIFF
--- a/internal/system/impl/sqlstore.go
+++ b/internal/system/impl/sqlstore.go
@@ -66,12 +66,21 @@ func (s *SystemSQLStoreService) GetTableMetadata(
 	}
 	store, ok := s.stores[chainID]
 	if !ok {
-		return sqlstore.TableMetadata{}, fmt.Errorf("chain id %d isn't supported in the validator", chainID)
+		return sqlstore.TableMetadata{
+			ExternalURL: fmt.Sprintf("%s/chain/%d/tables/%s", s.extURLPrefix, chainID, id),
+			Image:       s.emptyMetadataImage(),
+			Message:     "Chain isn't supported",
+		}, nil
 	}
 	table, err := store.GetTable(ctx, id)
 	if err != nil {
 		if !errors.Is(err, sql.ErrNoRows) {
-			return sqlstore.TableMetadata{}, fmt.Errorf("error fetching the table: %s", err)
+			log.Error().Err(err).Msg("error fetching the table")
+			return sqlstore.TableMetadata{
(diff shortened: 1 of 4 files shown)


pkg/sqlstore/impl/system/store.go
```diff
@@ -71,7 +71,7 @@ func (s *SystemStore) GetTable(ctx context.Context, id tableland.TableID) (sqlst
 		ID:      id.ToBigInt().Int64(),
 	})
 	if err != nil {
-		return sqlstore.Table{}, fmt.Errorf("failed to get the table: %s", err)
+		return sqlstore.Table{}, fmt.Errorf("failed to get the table: %w", err)
```
brunocalza, line 74: switch `%s` to `%w` so we can [check](https://github.com/tablelandnetwork/go-tableland/pull/219/files#diff-495917a0bcbd30f16dabee67340414731bcf65805956d9209a21da47227885e6R73) if the query returned rows or not

internal/system/impl/sqlstore.go
```diff
@@ -159,3 +169,9 @@ func (s *SystemSQLStoreService) getMetadataImage(chainID tableland.ChainID, tabl
 
 	return fmt.Sprintf("%s/%d/%s", uri, chainID, tableID)
 }
+
+func (s *SystemSQLStoreService) emptyMetadataImage() string {
```
brunocalza, line 173: calculates the value for field `image`

internal/system/impl/sqlstore.go
```diff
@@ -68,7 +70,15 @@ func (s *SystemSQLStoreService) GetTableMetadata(
 	}
 	table, err := store.GetTable(ctx, id)
 	if err != nil {
-		return sqlstore.TableMetadata{}, fmt.Errorf("error fetching the table: %s", err)
+		if !errors.Is(err, sql.ErrNoRows) {
+			return sqlstore.TableMetadata{}, fmt.Errorf("error fetching the table: %s", err)
```
brunocalza, line 74: If a database error occurs, we're not returning any table metadata and still returning an error. Kind of strange to return NFT metadata for a database error.

sanderpick (approved):
LGTM!

PR #227, opened 2022-08-18
Adds backuper component

brunocalza:
This component executes the backup of a database using SQLite API


pkg/backup/backuper.go
```diff
@@ -0,0 +1,326 @@
...
+	"path"
+	"time"
+
+	"github.com/mattn/go-sqlite3"
+
+	"github.com/pkg/errors"
+)
+
+// BackupFilenamePrefix is the prefix used in every backup file.
+const BackupFilenamePrefix = "tbl_backup"
```
brunocalza, line 17: This is used as a prefix in every backup file. [See](https://github.com/tablelandnetwork/go-tableland/pull/227/files#diff-c95ef053206f27ba2307970f3b1bfaa55fcd806df5e69e44123c010ccf796ac7R231).

pkg/backup/backuper.go
```diff
@@ -0,0 +1,326 @@
+package backup
```
brunocalza, line 1: Decided to use the word `backup` instead of `snapshot`. It looks more aligned with SQLite Backup API nomenclature.

brunocalza: If we decide to keep using the term backup, we should probably adjust our Notion docs to reflect that.

pkg/backup/backuper.go
```diff
@@ -0,0 +1,326 @@
...
+	"path"
+	"time"
+
+	"github.com/mattn/go-sqlite3"
+
+	"github.com/pkg/errors"
+)
+
+// BackupFilenamePrefix is the prefix used in every backup file.
+const BackupFilenamePrefix = "tbl_backup"
```
brunocalza, line 17: Every backup file has the prefix `tbl_backup`. [See](https://github.com/tablelandnetwork/go-tableland/pull/227/files#diff-c95ef053206f27ba2307970f3b1bfaa55fcd806df5e69e44123c010ccf796ac7R231).

pkg/backup/backuper.go
```diff
@@ -0,0 +1,326 @@
...
+
+var (
```
brunocalza, line 19: `open` and `newBackupFile` are global variable functions. This was done, so these function variables can be mocked in the tests. It will make sense when looking at the tests.

pkg/backup/backuper.go
```diff
@@ -0,0 +1,326 @@
...
+	open          = openDatabase
+	newBackupFile = createBackupFile
+)
+
+// Backuper is the process that executes the backup process.
+type Backuper struct {
```
brunocalza, line 25: The component that was added, responsible for backups.

The way it works:
- It receives in the constructor the database file from which the backup will be taken and a directory that has all the backup files
- It has only one method, the `Backup` that executes the whole logic and in the ends spits the result with some stats of the process

pkg/backup/backuper.go
```diff
@@ -0,0 +1,321 @@
...
+	}
+
+	b := &Backuper{
+		result: &BackupResult{
+			Dir: backupDir,
+		},
+		config: config,
+	}
+
+	if err := b.init(sourcePath, backupDir); err != nil {
```
brunocalza, line 47: The `init` creates a new backup file inside the `backupDir`, it opens both databases (source and backup), ping them to test connection. In other words, it prepares the `backuper`.

pkg/backup/backuper.go
```diff
@@ -0,0 +1,321 @@
...
+		panic("pruning not implemented")
+	}
+
+	b.result.SizeBeforeVacuum = backupSize
+	b.result.TableCount = tableCount
+	return b.result, nil
+}
+
+// init opens databases and ping them, then initializes variables.
+func (b *Backuper) init(sourcePath string, backupDir string) error {
```
brunocalza, line 115: I think this is pretty straightforward. I recommend taking a look at [openDatabase](https://github.com/tablelandnetwork/go-tableland/pull/227/files#diff-c95ef053206f27ba2307970f3b1bfaa55fcd806df5e69e44123c010ccf796ac7R212), which is the main implementation of `open`. Keep in mind that there's a mocked version of that in the tests. I'll touch on this in the tests.

pkg/backup/backuper.go
```diff
@@ -0,0 +1,321 @@
...
+	source, err := open(sourcePath)
+	if err != nil {
+		return errors.Errorf("opening source db: %s", err)
+	}
+
+	filename, timestamp, err := newBackupFile(backupDir, time.Now().UTC())
```
brunocalza, line 121: Check [createBackupFile](https://github.com/tablelandnetwork/go-tableland/pull/227/files#diff-c95ef053206f27ba2307970f3b1bfaa55fcd806df5e69e44123c010ccf796ac7R227). There's a mocked version of this in tests as well.

brunocalza: If you jumped to here, go back to [Backup method](https://github.com/tablelandnetwork/go-tableland/pull/227/files#diff-c95ef053206f27ba2307970f3b1bfaa55fcd806df5e69e44123c010ccf796ac7R55).

pkg/backup/backuper.go
```diff
@@ -0,0 +1,321 @@
...
+
+	if err := b.doBackup(connA, connB); err != nil {
+		return b.result, errors.Errorf("backup: %s", err)
+	}
+
+	// SQLite backup finishes
+	b.result.ElapsedTime = time.Since(startTime)
+
+	// count the number of tables for track record
+	tableCount, err := b.doTableCount(ctx, connB)
```
brunocalza, line 83: Not sure how useful collecting this is. I thought it would be nice to have this info for every backup to see how many tables are being created.

pkg/backup/backuper.go
```diff
@@ -0,0 +1,321 @@
...
+		return 0, 0, errors.Errorf("get file size: %s", err)
+	}
+
+	return size, time.Since(startTime), nil
+}
+
+func (b *Backuper) doTableCount(ctx context.Context, conn *sql.Conn) (int64, error) {
+	var tableCount int64
+	if err := conn.
+		QueryRowContext(ctx, "SELECT count(1) FROM sqlite_master WHERE type = 'table'").
```
brunocalza, line 195: I tried to use a query that is agnostic of tableland.

pkg/backup/backuper.go
```diff
@@ -0,0 +1,321 @@
...
+
+	if err := b.init(sourcePath, backupDir); err != nil {
+		return nil, err
+	}
+
+	return b, nil
+}
+
+// Backup creates a backup to a temporary in disk file.
+func (b *Backuper) Backup(ctx context.Context) (*BackupResult, error) {
```
brunocalza, line 55: It performs the backup into a single step. Along the way, it collects some metrics of the process. In the end, it executes more operations if enabled.

pkg/backup/backuper_test.go
```diff
@@ -0,0 +1,184 @@
...
+	"testing"
+	"time"
+
+	_ "github.com/mattn/go-sqlite3"
+	"github.com/stretchr/testify/require"
+)
+
+func TestBackuperDefault(t *testing.T) {
+	// substitutes the open function to a mocked version
+	open = func(uri string) (DB, error) {
```
brunocalza, line 16: This mocked implementation of `open` creates a `tempDatabase`. The `tempDatabase` is just regular `db` but it implements a special `Close` method that removes the database file. This is good for testing. It means we can create databases and destroy using the `DB` API.

pkg/backup/backuper_test.go
```diff
@@ -0,0 +1,184 @@
...
+		db, err := openDatabase(uri)
+		require.NoError(t, err)
+
+		return &tempDatabase{db}, nil
+	}
+
+	// substitutes the newBackupFile function to a mocked version
+	newBackupFile = func(dir string, _ time.Time) (string, time.Time, error) {
```
brunocalza, line 24: This mocked version is just to fix the timestamp of the backup, so we can test the filename.

pkg/backup/backuper_test.go
```diff
@@ -0,0 +1,184 @@
...
+		timestamp := time.Date(2009, 11, 17, 20, 34, 58, 651387237, time.UTC)
+		return createBackupFile(dir, timestamp)
+	}
+
+	controlDB := CreateControlDatabase(t)
```
brunocalza, line 29: We make use of a control database as ground truth. Check `control.sql`. 

That file creates a table, inserts 1000 rows and remove 500 odd rows to create space for vacuum. SQLite should execute `control.sql` deterministically, which means we can test the db size before and after vacuum.

pkg/backup/backuper.go
```diff
@@ -0,0 +1,321 @@
...
+	connB, err := b.result.Database.Conn(ctx)
+	if err != nil {
+		return b.result, errors.Errorf("getting backup db conn: %s", err)
+	}
+	defer func() {
+		_ = connB.Close()
+	}()
+
+	if err := b.doBackup(connA, connB); err != nil {
+		return b.result, errors.Errorf("backup: %s", err)
```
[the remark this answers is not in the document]
brunocalza: I don't return `BackupResult{}` because `b.result` contains the partially backuped file. Your comment below is a good suggestion and solves this issues by decouple the result from the information where the backup file is.

pkg/backup/backuper.go
```diff
@@ -0,0 +1,321 @@
...
+	}
+
+	// copies all pages in one single step
+	isDone, err := backup.Step(-1)
+	if err != nil {
+		return errors.Errorf("failed to perform the backup step: %s", err)
+	}
+	if !isDone {
+		return errors.New("backup is unexpectedly not done")
+	}
```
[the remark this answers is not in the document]
brunocalza: I'd wait a bit to see if the problems arises.

pkg/backup/backuper.go
```diff
@@ -0,0 +1,321 @@
...
+func createBackupFile(dir string, timestamp time.Time) (string, time.Time, error) {
+	filename := path.Join(dir, fmt.Sprintf("%s_%s.db", BackupFilenamePrefix, timestamp.Format(time.RFC3339)))
+	backupFile, err := os.Create(filename)
+	if err != nil {
+		return "", time.Time{}, errors.Errorf("creating backup file: %s", err)
+	}
+	if err := backupFile.Close(); err != nil {
+		return "", time.Time{}, errors.Errorf("closing backup file: %s", err)
+	}
+	return filename, timestamp, nil
```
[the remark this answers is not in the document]
brunocalza: Yeah. No need for this

PR #232, opened 2022-08-23
Add support for networks, direct write queries, validate

asutula:
This provides easier client configuration using hard coded information about the networks and contract addressees available. Also adds support for non-relayed write queries and rpc calls to validate write queries.

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -11,6 +11,7 @@ import (
 
 	"github.com/ethereum/go-ethereum/accounts/abi/bind"
 	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/ethclient"
 	"github.com/ethereum/go-ethereum/rpc"
 	"github.com/textileio/go-tableland/internal/tableland"
 	"github.com/textileio/go-tableland/pkg/nonce/impl"
@@ -19,22 +20,7 @@ import (
 	"github.com/textileio/go-tableland/pkg/wallet"
 )
 
-// Client is the Tableland client.
-type Client struct {
-	tblRPC      *rpc.Client
-	tblHTTP     *http.Client
-	tblContract *ethereum.Client
-	config      Config
-}
-
-// Config configures the Client.
-type Config struct {
-	TblAPIURL    string
-	EthBackend   bind.ContractBackend
-	ChainID      ChainID
-	ContractAddr common.Address
-	Wallet       *wallet.Wallet
-}
+var defaultNetwork = TestnetPolygonMumbai
 
 // TxnReceipt is a Tableland event processing receipt.
 type TxnReceipt struct {
@@ -63,9 +49,6 @@ func (tid TableID) ToBigInt() *big.Int {
 	return b
 }
 
-// ChainID is a supported EVM chain identifier.
-type ChainID int64
-
 // TableInfo summarizes information about a table.
 type TableInfo struct {
 	Controller string    `json:"controller"`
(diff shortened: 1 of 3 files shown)


pkg/client/client.go
```diff
@@ -86,25 +69,108 @@ func NewTableID(strID string) (TableID, error) {
...
+// NewClientContractBackend specifies a custom EVM backend to use.
+func NewClientContractBackend(backend bind.ContractBackend) NewClientOption {
+	return func(c *config) {
+		c.contractBackend = backend
+	}
+}
+
 // NewClient creates a new Client.
-func NewClient(ctx context.Context, config Config) (*Client, error) {
+func NewClient(ctx context.Context, wallet *wallet.Wallet, opts ...NewClientOption) (*Client, error) {
```
asutula, line 137: The client can now be created with options for which network, type of EVM backend to use (Alchemy, Infura, cusom) as well as a default setting for whether or not to relay writes.

pkg/client/client.go
```diff
@@ -222,30 +290,71 @@ func (c *Client) Read(ctx context.Context, query string) (string, error) {
...
+// Default behavior is false for main net EVM chains, true for all others.
+func WriteRelay(relay bool) WriteOption {
+	return func(wc *writeConfig) {
+		wc.relay = relay
 	}
+}
 
-	return res.Transaction.Hash, nil
+// Write initiates a write query, returning the txn hash.
+func (c *Client) Write(ctx context.Context, query string, opts ...WriteOption) (string, error) {
```
asutula, line 309: `Write` can now be called with an option to relay writes. If that option isn't provided, the `Client` default is used.

pkg/client/client.go
```diff
@@ -222,30 +290,71 @@ func (c *Client) Read(ctx context.Context, query string) (string, error) {
...
-
 	if err := c.tblRPC.CallContext(ctx, &res, "tableland_validateCreateTable", req); err != nil {
 		return "", fmt.Errorf("calling rpc validateCreateTable: %v", err)
 	}
-
 	return res.StructureHash, nil
 }
 
+// Validate validates a write query, returning the table id.
+func (c *Client) Validate(ctx context.Context, statement string) (TableID, error) {
```
asutula, line 344: Added support for this RPC because it is needed for making direct writes to validate the request and extract the table id from the statement.

pkg/client/networks.go
```diff
@@ -0,0 +1,163 @@
+package client
```
asutula, line 1: This file is really just the batteries to include with the `Client` to make it easy for users to select which network and EVM backend to use.

pkg/client/networks.go
```diff
@@ -0,0 +1,163 @@
...
+type Network string
+
+const (
+	// Testnet is the Tableland testnet network.
+	Testnet Network = "https://testnet.tableland.network"
+	// Staging is the Tableland staging network.
+	Staging Network = "https://staging.tableland.network"
+	// Localhost is the Tableland local network.
+	Localhost Network = "http://localhost:8080"
+)
```
asutula, line 17: This isn't sustainable, but it's the way it works now in the single validator world. Will revisit when we need to.

pkg/client/networks.go
```diff
@@ -0,0 +1,163 @@
...
+}
+
+// NetworkInfo is a info about a network supported by Talbleland.
+type NetworkInfo struct {
+	Network      Network
+	ChainID      ChainID
+	ContractAddr common.Address
+}
+
+var (
```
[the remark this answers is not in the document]
asutula: Yea i was thinking the same thing. Seems like a great idea.

pkg/client/networks.go
asutula: Good point. I added it for consistency with the JS SDK. Also maybe there will be a reason for a user to use it in a support situation with us. Will leave it for now for these reasons, but open to further changes here in the future if we decide they are needed.

sanderpick: The SDK includes all chains, but we don't advertise the staging ones on READMEs / docs. The CLI has a `tableland chains` command that list all supported chains, but staging ones are filtered out.

PR #236, opened 2022-08-23
Improved Client Chain Info

asutula:
Some quick improvements to the way the client was declaring chain information and utilizing it in the API.

After messaging with @sanderpick and learning more about the migration plan to Tableland mainnet, it became clear to me that my handing and naming of chain related data was overly complex and unnecessary. 

Also made the naming more consistent with the JS SDK by using the word "chain".

Also dropped declaration of `staging`-related things since that is for internal use. Users can still specify custom chain information if needing to connect to `staging` for support reasons.


pkg/client/chains.go
```diff
@@ -0,0 +1,133 @@
+package client
```
asutula, line 1: Renamed the file to `chains.go` from `networks.go`. Too bad it's showing it as a new file.

pkg/client/chains.go
```diff
@@ -0,0 +1,133 @@
...
+	Ethereum:       1,
+	Optimism:       10,
+	Polygon:        137,
+	EthereumGoerli: 5,
+	OptimismKovan:  69,
+	OptimismGoerli: 420,
+	ArbitrumGoerli: 421613,
+	PolygonMumbai:  80001,
+	Local:          31337,
+}
```
asutula, line 36: Got tired of polluting the top level namespace of the file with many `Chain` vars, so trying this approach to wrap them in a `struct`. Let me know if this is somehow offensive, but I like it I think.

pkg/client/chains.go
```diff
@@ -0,0 +1,133 @@
...
+
+// Chain is a info about a network supported by Talbleland.
+type Chain struct {
+	Endpoint     string
+	ID           ChainID
+	ContractAddr common.Address
+}
+
+// Chains is the connection info for all chains supported by Tableland.
+var Chains = struct {
```
asutula, line 46: Did the same namespacing with `struct` thing here too.

pkg/client/chains.go
```diff
@@ -0,0 +1,133 @@
+package client
+
+import (
+	"github.com/ethereum/go-ethereum/common"
+)
+
+const (
+	testnetURL = "https://testnet.tableland.network"
+	localURL   = "http://localhost:8080"
```
asutula, line 9: Once we launch a mainnet validator, will add it's URL here and update the mainnet EVM `Chain`s below to use that URL. Although, once there are many validators, we'll have to revisit this specifying of URL all together (as mentioned in my previous PR).

pkg/client/client.go
```diff
@@ -20,7 +20,7 @@ import (
 	"github.com/textileio/go-tableland/pkg/wallet"
 )
 
-var defaultNetwork = TestnetPolygonMumbai
+var defaultChain = Chains.PolygonMumbai
```
asutula, line 23: Nothing interesting in here, just updated for the refactor to `chains.go`.

sanderpick (approved):
very nice. looks good to me!

asutula:
@jsign, I added you here as a reviewer FYI. I went ahead merged this because it's basically a cleaning up of my PR you reviewed earlier today. In fact, it's probably more in line with a couple questions you were asking about earlier.

PR #241, opened 2022-08-24
Adds a configurable background routine that creates database backups

brunocalza:
# Summary

This PR adds a background scheduler that executes a database backup process every `X` minutes

# Details

## How the backup process works

The backup process called `Backuper` takes a backup of `SQLite` database file and stores it in a local directory relative to `--dir` (where database is stored).

The process makes use of the [SQLite Backup API](https://www.sqlite.org/c3ref/backup_finish.html) provided by [mattn/go-sqlite3](https://pkg.go.dev/github.com/mattn/go-sqlite3#SQLiteBackup). It is a full backup in a single step. Right now the database is small enough to not worry about locking and how long it takes. But an incremental backup approach may be needed when the database gets bigger.

## More operations besides backup

Besides the main backup of the database, the `Backuper` executes a `VACUUM` process in the backup file and compresses it with `zstd`. In the end, it prunes any excess of backup files according to configuration. Also, information such as elapsed time and database size are collected throughout the process.

## How scheduler works

The backup process is executed by a simple `Scheduler` that executes the backup routine every `X` minutes, where `X` is configurable. The scheduler is part of the validator.

## New configs

```
"Backup" : {
      "Enabled": true,                      // enables the backup scheduler to execute backups      
      "Dir": "backups",                     // directory relative to --dir where backups will be stored
      "Frequency": 240,                     // backup frequency in minutes
      "EnableVacuum": true,             
      "EnableCompression": true, 
      "Pruning" : {
        "Enabled": true,  // enables pruning
        "KeepFiles": 5    // pruning keeps at most `KeepFiles` backup files 
      }                 
}
```


cmd/api/config.go
```diff
@@ -44,6 +44,15 @@ type config struct {
 	Analytics struct {
 		FetchExtraBlockInfo bool `default:"false"`
 	}
+	Backup struct {
```
brunocalza, line 47: lmk if we should have all these configurable `Enable` flags or we can simply assume that the backup process will run with all operations for all validator

cmd/api/main.go
```diff
@@ -174,6 +175,26 @@ func main() {
 		}
 	}()
 
+	var backupScheduler *backup.Scheduler
+	if config.Backup.Enable {
+		backupScheduler, err = backup.NewScheduler(config.Backup.Frequency, backup.BackuperOptions{
```
brunocalza, line 180: If enabled starts the scheduler as a background routine

docker/deployed/staging/api/config.json
```diff
@@ -30,6 +30,15 @@
   "Analytics": {
     "FetchExtraBlockInfo": false
   },
+  "Backup" : {
+		"Enable": true,            
+		"Dir": "backups",               
+		"Frequency": 2,
```
brunocalza, line 36: 2 minutes in staging for testing

docker/deployed/testnet/api/config.json
```diff
@@ -31,6 +31,15 @@
     "Analytics": {
         "FetchExtraBlockInfo": true
     },
+    "Backup" : {
+		"Enable": true,            
+		"Dir": "backups",               
+		"Frequency": 240,
```
brunocalza, line 37: I'm thinking of running the backuper every 4h (240min)

pkg/backup/backuper.go
```diff
@@ -0,0 +1,356 @@
...
+const BackupFilenamePrefix = "tbl_backup"
+
+// Backuper is the process that executes the backup process.
+type Backuper struct {
+	sourcePath, dir string
+	source          DB
+	backup          DB
+	config          *Config
+
+	fileCreator func(string, time.Time) (string, error)
```
brunocalza, line 26: If you remember the previous implementation, it had global variables so some functions could be mocked. The problem with that is that we get race conditions when running tests in parallel. I switched to this approach.

So in tests, I will change how the backup file is created so I can test the name of the file.

pkg/backup/backuper.go
```diff
@@ -0,0 +1,356 @@
...
+
+	if err := os.MkdirAll(backupDir, 0o755); err != nil {
+		return nil, errors.Errorf("os mkdir all: %s", err)
+	}
+
+	b := &Backuper{
+		sourcePath:  sourcePath,
+		dir:         backupDir,
+		config:      config,
+		fileCreator: createBackupFile,
```
brunocalza, line 46: Sets the default implementation of `fileCreator`

pkg/backup/backuper.go
```diff
@@ -0,0 +1,356 @@
...
+	return nil
+}
+
+// Backup creates a backup to a file in disk.
+// Multiple calls to Backup can be perfomed.
+// But the file in which backup occurs was already created, which means the backup always occur at the same file.
+// This can be used to perform retries in case of errors.
+func (b *Backuper) Backup(ctx context.Context) (_ BackupResult, err error) {
+	defer func() {
+		if err != nil {
```
brunocalza, line 81: If any error occurs in the backup process we'll delete the temporary backup file

pkg/backup/backuper.go
```diff
@@ -0,0 +1,356 @@
...
+	}
+
+	if b.config.Vacuum {
+		backupResult.SizeAfterVacuum, backupResult.VacuumElapsedTime, err = b.doVacuum(ctx, connB)
+		if err != nil {
+			return BackupResult{}, errors.Errorf("do vacuum: %s", err)
+		}
+	}
+
+	// closes connections because we don't need any more access to dbs
```
brunocalza, line 122: This is important so we don't have `-wal` and `-shm` files hanging around

pkg/backup/pruner.go
```diff
@@ -0,0 +1,65 @@
...
+}
+
+func readBackupFiles(dir string) ([]fs.FileInfo, error) {
+	files, err := os.ReadDir(dir)
+	if err != nil {
+		return []fs.FileInfo{}, fmt.Errorf("read dir: %s", err)
+	}
+
+	backupFiles := []fs.FileInfo{}
+	for _, f := range files {
```
brunocalza, line 46: This loop filters to consider only backup files, that is files that start with `tbl_backup` and end with `.db` or `.db.gz`.

This is to accidentally not remove any other kind of file.

pkg/backup/pruner.go
```diff
@@ -0,0 +1,65 @@
...
+	for _, file := range toBeRemoved {
+		if err := os.Remove(path.Join(dir, file.Name())); err != nil {
+			return errors.Errorf("os remove: %s", err)
+		}
+	}
+
+	return nil
+}
+
+func readBackupFiles(dir string) ([]fs.FileInfo, error) {
```
brunocalza, line 39: Get a list of all backup files sorted by mod time.

pkg/backup/scheduler.go
```diff
@@ -0,0 +1,112 @@
...
+
+// BackuperOptions options needed to instantiate a backuper.
+type BackuperOptions struct {
+	SourcePath, BackupDir string
+	Opts                  []Option
+}
+
+// NewScheduler creates a new backup scheduler.
+func NewScheduler(frequency int, opts BackuperOptions, notify bool) (*Scheduler, error) {
+	if frequency < 1 || frequency >= 1440 {
```
brunocalza, line 37: It should be greater than 1 minute and less than a day

pkg/backup/scheduler.go
```diff
@@ -0,0 +1,112 @@
...
+	"github.com/pkg/errors"
+	logger "github.com/rs/zerolog/log"
+)
+
+var log = logger.With().Str("component", "backup").Logger()
+
+// Scheduler executes backups at a regular interval.
+type Scheduler struct {
+	Frequency      int // in minutes
+	NotificationCh chan bool
```
brunocalza, line 18: The `NotificationCh` is used to notify them that the backup process is finished for the client. This is used only in the test right now but can be leveraged by other occasions in the future.

pkg/backup/scheduler.go
```diff
@@ -0,0 +1,112 @@
...
+
+	return &Scheduler{
+		Frequency:      frequency,
+		NotificationCh: make(chan bool),
+
+		notify:   notify,
+		backuper: backuper,
+		close:    make(chan struct{}),
+
+		tickerInterval: time.Minute,
```
brunocalza, line 54: The scheduler ticker ticks every minute

pkg/backup/scheduler.go
```diff
@@ -0,0 +1,112 @@
...
+	}, nil
+}
+
+// Run starts the scheduler and listens for a shutdown call.
+func (s *Scheduler) Run() {
+	log.Info().Msg("starting backup scheduler")
+
+	// wait until next interval to start
+	now, interval := time.Now(), time.Duration(s.Frequency)*s.tickerInterval
```
brunocalza, line 63: I did a very simple implementation of scheduling. It figures out that the next time the backup should be run according to the configurable frequency. It will wait until then to execute the first backup. 

After that, it executes the backup and waits until the next time again.

pkg/backup/scheduler.go
```diff
@@ -0,0 +1,112 @@
...
+	if err := s.backuper.Init(); err != nil {
+		log.Error().Err(err).Msg("initializing backuper")
+		return
+	}
+	result, err := s.backuper.Backup(context.Background())
+	if err != nil {
+		log.Error().Err(err).Msg("backup failed")
+		return
+	}
+	log.Info().
```
brunocalza, line 100: I'm relying on logs right now to get stats from the backup process. We could use otel if that's the case. I did not add it because it adds so much clutter to the code, that I prefer waiting to see if it's really needed.

pkg/backup/scheduler.go
brunocalza: `wait := now.Round(interval).Sub(now)` may be cleaner

cmd/api/config.go
```diff
@@ -44,6 +44,15 @@ type config struct {
...
+		Enable            bool   `default:"true"`
+		Dir               string `default:"backups"` // relative to dir path config (e.g. ${HOME}/.tableland/backups )
+		Frequency         int    `default:"120"`     // in minutes
+		EnableVacuum      bool   `default:"true"`
+		EnableCompression bool   `default:"true"`
+		EnablePruning     bool   `default:"true"`
+		KeepFiles         int    `default:"5"` // number of files to keep
```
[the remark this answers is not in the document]
brunocalza: Sounds good!

pkg/backup/backuper.go
```diff
@@ -0,0 +1,356 @@
...
+
+	backup, err := open(filename)
+	if err != nil {
+		return errors.Errorf("opening backup db: %s", err)
+	}
+
+	b.source = source
+	b.backup = backup
+	return nil
+}
```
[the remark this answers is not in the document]
brunocalza: This is a good point. I tried so many APIs for Backuper that I kind of forgot why it end up this way. I'll see if making `init` private and calling it inside backup will work with current tests and clients.

pkg/backup/backuper.go
```diff
@@ -0,0 +1,356 @@
...
+
+// Backup creates a backup to a file in disk.
+// Multiple calls to Backup can be perfomed.
+// But the file in which backup occurs was already created, which means the backup always occur at the same file.
+// This can be used to perform retries in case of errors.
```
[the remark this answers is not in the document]
brunocalza: Yeah. You're right, multiple serial calls. There's even a test for that. The idea is that after you create a backuper you can reuse it (not concurrently) to generate new backups.

pkg/backup/compressor.go
```diff
@@ -0,0 +1,34 @@
...
+		return "", fmt.Errorf("gzip write: %s", err)
+	}
+	if err := w.Close(); err != nil {
+		return "", fmt.Errorf("close: %s", err)
+	}
+
+	newFilepath := fmt.Sprintf("%s.gz", filepath)
+	if err := os.WriteFile(newFilepath, b.Bytes(), 0o755); err != nil {
+		return "", fmt.Errorf("write file: %s", err)
+	}
```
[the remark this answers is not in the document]
brunocalza: Makes sense. I thought this would be a problem, but I wasn't sure if the `gzip` compression would work in stream mode (if it has to read the whole file beforehand or not). Since I did not find with a simple Google search, I went with the whole file in memory approach.

pkg/backup/pruner.go
brunocalza: This is a good point. Specially if we add a pruning CLI command that can be run at the same time as the background routine.

pkg/backup/scheduler.go
```diff
@@ -0,0 +1,113 @@
...
+		return nil, errors.New("interval should be in [1,1440)")
+	}
+
+	backuper, err := NewBackuper(opts.SourcePath, opts.BackupDir, opts.Opts...)
+	if err != nil {
+		return nil, fmt.Errorf("new backuper: %s", err)
+	}
+
+	return &Scheduler{
+		Frequency:      frequency,
```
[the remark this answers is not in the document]
brunocalza: Good suggestion! Yeah. I change the ticker from minutes to seconds to test the scheduler. It's hacky, but any other approach to test would add too much code.

pkg/backup/scheduler.go
```diff
@@ -0,0 +1,113 @@
...
+		select {
+		case <-s.close:
+			log.Info().Msg("closing backup scheduler")
+			return
+		case <-time.After(wait):
+			startTime := time.Now()
+			s.backup()
+			if s.notify {
+				s.NotificationCh <- true
+			}
```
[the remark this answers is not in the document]
brunocalza: I did not spend too much time thinking into this, actually. The current behavior is that it just notifies that the process finished, without saying if it was with error or not. This was done to make test easier and avoid `require.Eventually` that may give flaky tests. Maybe a better approach for now would be to just return an error in `s.backup` and put a `false` or `true` depending on or error. (Or return `BackupResult`)

pkg/backup/scheduler.go
```diff
@@ -0,0 +1,113 @@
...
+			wait = time.Duration(s.Frequency)*s.tickerInterval - time.Since(startTime)
+		}
+	}
+}
+
+// Shutdown gracefully shutdowns the scheduler.
+func (s *Scheduler) Shutdown() {
+	s.closeOnce.Do(func() {
+		s.close <- struct{}{}
```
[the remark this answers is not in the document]
brunocalza: Man, this is something that bugged me. I thought so, but could not make the shutdown work properly without this. Maybe I'm missing something. I'll test again to make sure.

pkg/backup/scheduler.go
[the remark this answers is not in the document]
brunocalza: @jsign I'm using `tickerFrequency` suggestion, it worked

pkg/backup/scheduler.go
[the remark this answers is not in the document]
brunocalza: @jsign i will change `NotificationCh` to `chan error` to give more signal to clients, but I won't change the L77 behavior. sounds good for now

cmd/api/config.go
```diff
@@ -44,6 +44,17 @@ type config struct {
 	Analytics struct {
 		FetchExtraBlockInfo bool `default:"false"`
 	}
+	Backup struct {
+		Enabled           bool   `default:"true"`
+		Dir               string `default:"backups"` // relative to dir path config (e.g. ${HOME}/.tableland/backups )
+		Frequency         int    `default:"120"`     // in minutes
+		EnableVacuum      bool   `default:"true"`
+		EnableCompression bool   `default:"true"`
+		Pruning           struct {
```
brunocalza, line 53: Better config structure by grouping the pruning related config

pkg/backup/backuper.go
```diff
@@ -0,0 +1,354 @@
...
+// Backup creates a backup to a file in disk.
+// Multiple serial calls to Backup can be perfomed. This can be used to perform retries in case of errors.
+func (b *Backuper) Backup(ctx context.Context) (_ BackupResult, err error) {
+	defer func() {
+		if err != nil {
+			_ = os.Remove(b.backup.Path())
+		}
+	}()
+
+	timestamp, err := b.init()
```
brunocalza, line 61: `init` is now private and used inside Backup. This means that the backup file is created at the same time the process will be execute. Much cleaner API for clients.

pkg/backup/compressor.go
```diff
@@ -0,0 +1,61 @@
...
+
+	"github.com/klauspost/compress/zstd"
+
+	"github.com/pkg/errors"
+	"golang.org/x/sync/errgroup"
+)
+
+const extension = "zst"
+
+// Compress compresses a file using zstd.
```
brunocalza, line 17: Using `zstd` as the compression algorithm and doing the process in stream mode to avoid reading the whole file into memory

pkg/backup/scheduler.go
brunocalza: Added the `mLastExecution` OTEL metric. It keeps track of the successful backup file timestamps (for clarification, not the time the backup finished)

pkg/backup/scheduler.go
```diff
@@ -0,0 +1,141 @@
...
+	backuper, err := NewBackuper(opts.SourcePath, opts.BackupDir, opts.Opts...)
+	if err != nil {
+		return nil, fmt.Errorf("new backuper: %s", err)
+	}
+
+	s := &Scheduler{
+		NotificationCh:  make(chan error),
+		notify:          notify,
+		backuper:        backuper,
+		tickerFrequency: time.Duration(frequency) * time.Minute,
```
brunocalza, line 54: Using `tickerFrequency` as the only property to calculate how long to wait for the next backup run. In tests, we change this to a higher frequency to execute the test faster

jsign (approved):
LGTM!

PR #250, opened 2022-08-30
Parser update

brunocalza:
Updates SQL parser to the most recent version which enforces the following rule from [SQL spec ](https://docs.tableland.xyz/sql-specification#9e191076e01b4ac5a24d684060b9dc49)in the parser: Each table in Tableland may have at most one PRIMARY KEY.

The error behavior of the parser changed a bit as explained in the [PR](https://github.com/tablelandnetwork/go-sqlparser/pull/11). So, there are some changes in this PR to reflect the change in that behavior.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -20,7 +20,7 @@ require (
 	github.com/spf13/cobra v1.5.0
 	github.com/spruceid/siwe-go v0.2.1-0.20220711143404-817846826282
 	github.com/stretchr/testify v1.8.0
-	github.com/tablelandnetwork/sqlparser v0.0.0-20220725134734-b79ca3e804c8
+	github.com/tablelandnetwork/sqlparser v0.0.0-20220830151425-aec4df78e4c8
 	github.com/textileio/cli v1.0.2
 	go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp v0.32.0
 	go.opentelemetry.io/contrib/instrumentation/runtime v0.34.0
--- a/pkg/parsing/impl/validator.go
+++ b/pkg/parsing/impl/validator.go
@@ -48,7 +48,7 @@ func New(systemTablePrefixes []string, opts ...parsing.Option) (parsing.SQLValid
 func (pp *QueryValidator) ValidateCreateTable(query string, chainID tableland.ChainID) (parsing.CreateStmt, error) {
 	ast, err := sqlparser.Parse(query)
 	if err != nil {
-		return nil, &parsing.ErrInvalidSyntax{InternalError: err}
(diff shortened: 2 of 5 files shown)


go.mod
```diff
@@ -20,7 +20,7 @@ require (
 	github.com/spf13/cobra v1.5.0
 	github.com/spruceid/siwe-go v0.2.1-0.20220711143404-817846826282
 	github.com/stretchr/testify v1.8.0
-	github.com/tablelandnetwork/sqlparser v0.0.0-20220725134734-b79ca3e804c8
+	github.com/tablelandnetwork/sqlparser v0.0.0-20220830151425-aec4df78e4c8
```
brunocalza, line 23: latest `main` commit

pkg/parsing/impl/validator.go
```diff
@@ -48,7 +48,7 @@ func New(systemTablePrefixes []string, opts ...parsing.Option) (parsing.SQLValid
 func (pp *QueryValidator) ValidateCreateTable(query string, chainID tableland.ChainID) (parsing.CreateStmt, error) {
 	ast, err := sqlparser.Parse(query)
 	if err != nil {
-		return nil, &parsing.ErrInvalidSyntax{InternalError: err}
```
brunocalza, line 51: I removed the `parsing.ErrInvalidSyntax`, now we just return a wrapper error since `err` can be an invalid syntax or another error such as multiple primary keys.

pkg/parsing/impl/validator_test.go
```diff
@@ -896,8 +896,8 @@ func newParser(t *testing.T, prefixes []string, opts ...parsing.Option) parsing.
 }
 
 // Helpers to have a pointer to pointer for generic test-case running.
-func ptr2ErrInvalidSyntax() **parsing.ErrInvalidSyntax {
-	var e *parsing.ErrInvalidSyntax
+func ptr2ErrInvalidSyntax() **sqlparser.ErrSyntaxError {
```
brunocalza, line 899: Some tests check for errors in syntax, so this change was necessary so tests could still pass.

pkg/parsing/impl/validator.go
```diff
@@ -60,10 +60,6 @@ func (pp *QueryValidator) ValidateCreateTable(query string, chainID tableland.Ch
 		return nil, &parsing.ErrNoTopLevelCreate{}
 	}
 
-	if ast.Errors[0] != nil {
```
brunocalza, line 63: No more need to check `ast.Errors`, the need for this additional check was confusing

jsign (approved):
:heavy_check_mark:

PR #253, opened 2022-08-30
APIs Refactor

asutula:
This PR is conceptually a simple one, but with lots of naming updates in the code, thus the high "files changed" count. Files that don't contain any comments can just be considered as having simple naming updates to accommodate the refactor.

This PR doesn't functionally change anything about `go-tableland`, but sets the stage for some upcoming changes to support improved read query parameters in both the REST and JSON RPC APIs, as well as making it easy to add or remove any existing or new API service (for example, we may choose to not have a JSON RPC API).

The main change here is to free the `tableland.Tableland` interface and its implementation `TablelandMesa` from JSON RPC duties. The goal is to make it the core API that can be used by any other component of `go-tableland`, using whatever expressive API makes sense, and not be forced to conform to JSON RPC rules. 

The `tableland.Tableland` interface was updated to reflect this goal (not using `Request` and `Response` structs for every function), and additional, more fluent Go interactions will be added in an upcoming PR to support the new read query parameters.

The REST API `UserController` was updated to use the new core API.

A new `RPCService` was created to actually provide the JSON RPC service. This service uses the core API in its implementation.


internal/router/controllers/user.go
```diff
@@ -320,23 +321,20 @@ func (c *UserController) runReadRequest(
 	stm string,
 	rw http.ResponseWriter,
 ) (*sqlstore.UserRows, bool) {
-	req := tableland.RunReadQueryRequest{
-		Statement: stm,
-	}
-	res, err := c.runner.RunReadQuery(ctx, req)
+	res, err := c.runner.RunReadQuery(ctx, stm)
```
asutula, line 324: This controller was simply updated to use the new `tableland.Tableland` API.

internal/router/router.go
```diff
@@ -35,13 +36,15 @@ func ConfiguredRouter(
 	}
 
 	mesaService := impl.NewTablelandMesa(parser, instrUserStore, chainStacks)
```
asutula, line 38: This is now our core API. No JSON RPC duties.

internal/router/router.go
```diff
@@ -35,13 +36,15 @@ func ConfiguredRouter(
...
-	userController := controllers.NewUserController(mesaService)
 	mesaService, err = impl.NewInstrumentedTablelandMesa(mesaService)
 	if err != nil {
 		log.Fatal().Err(err).Msg("instrumenting mesa")
 	}
+	rpcService := rpcservice.NewRPCService(mesaService)
+	userController := controllers.NewUserController(mesaService)
```
asutula, line 44: The REST API and new RPC service being initialized with the core API.

internal/router/router.go
```diff
@@ -35,13 +36,15 @@ func ConfiguredRouter(
...
+
 	server := rpc.NewServer()
-	if err := server.RegisterName("tableland", mesaService); err != nil {
+	if err := server.RegisterName("tableland", rpcService); err != nil {
```
asutula, line 47: Now using the RPC service here.

internal/router/rpcservice/rpcservice.go
```diff
@@ -0,0 +1,191 @@
...
+type RelayWriteQueryRequest struct {
+	Statement string `json:"statement"`
+}
+
+// RelayWriteQueryResponse is a RelayWriteQuery response.
+type RelayWriteQueryResponse struct {
+	Transaction struct {
+		Hash string `json:"hash"`
+	} `json:"tx"`
+}
```
asutula, line 22: All of the request/response types now exist in here, no longer in mesa.

internal/router/rpcservice/rpcservice.go
```diff
@@ -0,0 +1,191 @@
...
+		return SetControllerResponse{}, fmt.Errorf("parsing token ID: %v", err)
+	}
+	txn, err := rs.tbl.SetController(ctx, common.HexToAddress(req.Controller), tableID)
+	if err != nil {
+		return SetControllerResponse{}, fmt.Errorf("calling SetController: %v", err)
+	}
+	ret := SetControllerResponse{}
+	ret.Transaction.Hash = txn.Hash().Hex()
+	return ret, nil
+}
```
asutula, line 191: Pretty simple wrapping of the core API for now, but the point is that there is now this layer of indirection where the JSON RPC service can provide additional logic or different behavior. This will be exercised in the next PR.

internal/tableland/impl/mesa.go
```diff
@@ -35,150 +36,131 @@ func NewTablelandMesa(
 
 // ValidateCreateTable allows to validate a CREATE TABLE statement and also return the structure hash of it.
 // This RPC method is stateless.
-func (t *TablelandMesa) ValidateCreateTable(
-	ctx context.Context,
-	req tableland.ValidateCreateTableRequest,
-) (tableland.ValidateCreateTableResponse, error) {
+func (t *TablelandMesa) ValidateCreateTable(ctx context.Context, statement string) (string, error) {
```
asutula, line 39: Obviously some real changes in here, but will be easier to comment on in the interface declaration for `tableland.Tableland`.

internal/tableland/impl/mesa_instrumented.go
```diff
@@ -98,28 +91,29 @@ func (t *InstrumentedTablelandMesa) RelayWriteQuery(
...
+) (tables.Transaction, error) {
 	start := time.Now()
-	resp, err := t.tableland.SetController(ctx, req)
+	resp, err := t.tableland.SetController(ctx, controller, tableID)
 	latency := time.Since(start).Milliseconds()
 
-	controller, _ := ctx.Value(middlewares.ContextKeyAddress).(string)
 	chainID, _ := ctx.Value(middlewares.ContextKeyChainID).(tableland.ChainID)
-	t.record(ctx, recordData{"SetController", controller, "", err == nil, latency, chainID})
+	t.record(ctx, recordData{"SetController", controller.Hex(), "", err == nil, latency, chainID})
```
asutula, line 116: I think there was a little bug here. In the old code, it was pulling an address out of the context using `middlewares.ContextKeyAddress` then using that address to record as the `controller`. But the address in the context is actually the `caller`. The new code just uses the provided `controller` param to record the controller.

internal/tableland/tableland.go
```diff
@@ -50,102 +23,27 @@ type TxnReceipt struct {
...
-type SetControllerResponse struct {
-	Transaction struct {
-		Hash string `json:"hash"`
-	} `json:"tx"`
-}
-
 // SQLRunner defines the run SQL interface of Tableland.
 type SQLRunner interface {
-	RunReadQuery(context.Context, RunReadQueryRequest) (RunReadQueryResponse, error)
+	RunReadQuery(context.Context, string) (interface{}, error)
```
asutula, line 28: Example of a more fluent Go API here, no need for the request/response types.

internal/tableland/tableland.go
```diff
@@ -50,102 +23,27 @@ type TxnReceipt struct {
...
-// NewTableIDFromInt64 returns a TableID from a int64.
-func NewTableIDFromInt64(intID int64) (TableID, error) {
-	tableID := &big.Int{}
-	tableID.SetInt64(intID)
-	return TableID(*tableID), nil
+	ValidateCreateTable(context.Context, string) (string, error)
+	ValidateWriteQuery(context.Context, string) (tables.TableID, error)
+	RelayWriteQuery(context.Context, string) (tables.Transaction, error)
+	GetReceipt(context.Context, string) (bool, *TxnReceipt, error)
+	SetController(context.Context, common.Address, tables.TableID) (tables.Transaction, error)
```
asutula, line 38: Also, since we're not needing to provide JSON RPC conformity, we can use more fluent types like `TableID`, `Transaction`, and `TxnReceipt` in the API.

pkg/backup/testing.go
```diff
@@ -12,7 +11,7 @@ import (
 func createControlDatabase(t *testing.T) DB {
 	t.Helper()
 
-	f, err := ioutil.TempFile(t.TempDir(), "control_*.db")
+	f, err := os.CreateTemp(t.TempDir(), "control_*.db")
```
asutula, line 14: Had some lint warnings about this here and similar in a couple other places.

pkg/tables/tables.go
```diff
@@ -2,12 +2,48 @@ package tables
...
 	"context"
+	"fmt"
 	"math/big"
 
 	"github.com/ethereum/go-ethereum/common"
-	"github.com/textileio/go-tableland/internal/tableland"
 )
 
+// TableID is the ID of a Table.
+type TableID big.Int
```
asutula, line 12: I moved this type here from `tableland/tableland.go` to avoid an import cycle between that file and this one. It seemed arbitrary to me where it is defined, either place makes sense, so this was the simplest work around.

In retrospect, I'm seeing that it is what caused so many file changes making the PR bigger. Sorry about that.

internal/tableland/impl/mesa.go
```diff
@@ -35,150 +36,131 @@ func NewTablelandMesa(
...
+func (t *TablelandMesa) ValidateWriteQuery(ctx context.Context, statement string) (tables.TableID, error) {
 	ctxController := ctx.Value(middlewares.ContextKeyAddress)
 	controller, ok := ctxController.(string)
 	if !ok || controller == "" {
-		return tableland.ValidateWriteQueryResponse{}, errors.New("no controller address found in context")
+		return tables.TableID{}, errors.New("no controller address found in context")
 	}
 
 	ctxChainID := ctx.Value(middlewares.ContextKeyChainID)
 	chainID, ok := ctxChainID.(tableland.ChainID)
```
asutula, line 61: I was considering making the controller address and chain ID explicit parameters to this func and one other that does the same thing. This would feel more correct to me for a "core API", no hidden surprises. I left it as is to get some input from reviewers. What do you think?

jsign: Agree, I think it is a good idea.

I think the `ctx` for storing values is usually a slippery slope and too opaque. Moreover, I think the `ctx` use here comes from the fact that this service was tied to RPC, and usually these values come from middleware that can only send them injected in `ctx`.

Technically speaking, if we think about the CoreAPI as untangled from RPCs, is up to the caller to use whatever mechanism they can, and unpack that complexity there while passing "plain arguments" at this layer.

I'm fine too with creating a `TblContext` struct if we think that's worth it to avoid parameter bloat. I don't think that's the case yet since we have about two usual parameters (`ChainID` and `Caller`). No strong opinions.

internal/tableland/impl/mesa.go
```diff
@@ -35,150 +36,131 @@ func NewTablelandMesa(
...
 	}
 
 	queryResult, err := t.runSelect(ctx, readStmt)
 	if err != nil {
-		return tableland.RunReadQueryResponse{}, fmt.Errorf("running read statement: %s", err)
+		return nil, fmt.Errorf("running read statement: %s", err)
 	}
-	return tableland.RunReadQueryResponse{Result: queryResult}, nil
+	return queryResult, nil
 }
```
jsign, line 141: @asutula, do you think it may be possible to move the formatting modes at this layer in your next PR? Or would that still be a particular feature of the RPC flavor (REST)?

I'd suspect the latter since formatting things are more related to RPC layers than CoreAPIs. If that's the case, this was the what I was referring to in the tension about duplicate work since not every "API feature" can be generalized.

In any case, I don't know yet what would be the ultimate conclusion. If we end up dropping JSON-RPC, then it might be irrelevant.

internal/tableland/impl/mesa_instrumented.go
jsign: I think it kind of depends on if we want to record who made the call versus which is the target new controller. If I'd try to make the interpretation today, I think I'd lean to what you did but would still have doubts.

In any case, we should really remove that `controller` recording in metrics. That leads to high cardinality dimensions that aren't nice for Prometheus when we have more volume. So probably in the medium term, it will be removed, so the dilemma will be solved. ha

internal/tableland/tableland.go
jsign: I don't recall from memory, but can this interface be defined in the caller site?

I think it is just the caller trying to strip down the need for the mesa dependency to call `RunReadQuery`; I'm not sure if it makes sense to have that in such a general package as it's here.

pkg/eventprocessor/impl/executor/impl/blockscope.go
```diff
@@ -60,10 +60,10 @@ func newBlockScope(
 //
 // If the events execution is successful, it returns the result.
 // If the events execution isn't successful, we have two cases:
-// 1) If caused by controlled error (e.g: invalid SQL), it will return a (res, nil) where
-//    res contains the error message.
-// 2) If caused by uncontrolled error (e.g: can't access the DB), it returns ({}, err). The caller should retry
-//    executing this transaction events when the underlying infrastructure problem is solved.
+//  1. If caused by controlled error (e.g: invalid SQL), it will return a (res, nil) where
+//     res contains the error message.
+//  2. If caused by uncontrolled error (e.g: can't access the DB), it returns ({}, err). The caller should retry
```
jsign, line 65: Was this surfaced by some lint? (Agree with the change, just trying to force these things as `make lint` if possible so we have this as an automatic rule)

pkg/tables/tables.go
jsign: I think it's fine. I might have a better opinion while being hands-on with the code in the future with the new organization.

jsign (approved):
LGTM! I think this refactor goes in the right direction :+1: 

From what I understood, this shouldn't have any impact on the outside-world visible behavior of our API, but there's still some chance that some mapping isn't 100% correct in the refactor.
I don't know if you've already done some test with `[stagingdeploy]`, but in any case, we should keep a close eye on whenever the PR is merged if someone gets any strange behavior.

internal/tableland/impl/mesa.go
brunocalza: My personal view is that formatting modes should be out of the Core API. The core API should be agnostic on how its internal representation will be presented to the external world. Maybe we'll have a client that for some reason need a slightly different format output (i know we're aiming for consistency across all clients but sometimes a client has specific needs). 

The way I'd design is by having "presenter" or "formatter" components that would receive the Tableland Core Data Structure and format in any way it's relevant. The clients that need consistency would call the same presenter component. 

Hopefully, that makes sense.

internal/tableland/tableland.go
brunocalza: I'd vote to remove this interface if possible

pkg/backup/testing.go
brunocalza: yeah. `ioutil` pkg is deprecated. I tried to find the equivalent in the `os` but did not find it at the time. are you using a different linter or how did you set it up? I did not see any warnings

brunocalza (approved):
Very nice refactor 👏

internal/tableland/impl/mesa.go
jsign: Makes sense to me.

internal/tableland/impl/mesa_instrumented.go
asutula: Ok will leave as is.

pkg/backup/testing.go
asutula: You know, I'm not sure how my vscode linter is picking up these extra things that `make lint` isn't. Could be part of vscode running gofumpt for formatting, or the fact that i'm using go 1.19 now? Will have to research this more to make sure my vscode and `make lint` do the same thing, and that we all get those settings coming along with the repo.

pkg/tables/tables.go
asutula: Cool, will leave it like this for now.

pkg/eventprocessor/impl/executor/impl/blockscope.go
asutula: Yea, somehow my vscode go linter or gofumpt formatting did this. Will try to get our `make lint` flagging these things too, not sure why it's not. Pretty sure this was done by gofumpt on file save in vscode.

internal/tableland/impl/mesa.go
asutula: Yea, what @brunocalza described is what i was planning. Basically, the core api will return `UserRows`, then we'll have the presenter layer (this could just be a function) transform that according to the new output format params provided by the request (both REST and JSON RPC). There is one case where the JSON RPC layer will have some special behavior (when the user asks for `jsonl` output) where it will need to either encode the result as base64 string, or error if the `jsonl` is for more than one object being returned. This setup we're talking about will support that kind of flexibility on a per API basis.

asutula: I think we're all on the same page here, so going to resolve this convo.

internal/tableland/impl/mesa.go
asutula: Ok those params are now explicit. Was able to clean up a bit of cruft in the tests because of this.

internal/tableland/tableland.go
asutula: I moved this interface to the caller site.

pkg/backup/testing.go
asutula: So, this one is showing up for me when running `make lint`.

```
> make lint
go run github.com/golangci/golangci-lint/cmd/golangci-lint@v1.46.2 run
pkg/backup/testing.go:4:2: SA1019: package io/ioutil is deprecated: As of Go 1.16, the same functionality is now provided by package io or package os, and those implementations should be preferred in new code. See the specific function documentation for details. (staticcheck)
        "io/ioutil"
        ^
exit status 1
make: *** [lint] Error 1
```

Do you not see that? I am running Go 1.19.

pkg/eventprocessor/impl/executor/impl/blockscope.go
asutula: Interesting. This was fixed because of my auto format on save with the formatter set to gofumpt. But, if i revert this change using vi (no format on save) and run `make lint`, it does fail linting:

```
> make lint
go run github.com/golangci/golangci-lint/cmd/golangci-lint@v1.46.2 run
pkg/eventprocessor/impl/executor/impl/blockscope.go:63: File is not `gofumpt`-ed (gofumpt)
//  1) If caused by controlled error (e.g: invalid SQL), it will return a (res, nil) where
pkg/eventprocessor/impl/executor/impl/blockscope.go:65: File is not `gofumpt`-ed (gofumpt)
//  2) If caused by uncontrolled error (e.g: can't access the DB), it returns ({}, err). The caller should retry
exit status 1
make: *** [lint] Error 1
```

I'm running Go 1.19, could that have something to do with it? I can't think of how else this wouldn't fail in CI and on your machine.

jsign: @asutula , I confirmed that's the case. With go1.17.2, the lint doesn't fail, but if I update to go1.19 it does.
Maybe in go.1.19 something changed in the stdlib reg parsing files and the AST changed, thus gofumpt can detect something new (without it's code being changed).

We should change in `.github/workflows/review.yaml` the version from go1.17 to go1.19. We're actually using go1.19 for tests and Dockerfiles.

asutula: Ok nice, will do that. I also noticed there is an updated version of golangci-lint that detects a few new issues about missing package comments. Will look at updating that as well.

asutula: The package comment requirements coming from `golangci-lint` `revive` plugin when upgrading to the latest `golangci-lint` are a little confusing to me at the moment. Will leave `golangci-lint` at the version we're already using and just update the go version. Can revisit that upgrade later.

sanderpick:
Very nice! Great to see this.

PR #255, opened 2022-08-30
executor: improve closing

jsign:
The `Executor` had two missing things:
1. Close the db when `Close()` is called. (I wonder if this contributed to WAL prunning checkpointing delays)
2. There was a subtle race between `Close()` and getting a lock to enforce serial execution (will be explained in comments).

### DIFF
--- a/pkg/eventprocessor/impl/executor/impl/executor.go
+++ b/pkg/eventprocessor/impl/executor/impl/executor.go
@@ -5,6 +5,7 @@ import (
 	"database/sql"
 	"errors"
 	"fmt"
+	"sync"
 
 	"github.com/XSAM/otelsql"
 	"github.com/mattn/go-sqlite3"
@@ -26,6 +27,9 @@ type Executor struct {
 
 	chainID          tableland.ChainID
 	maxTableRowCount int
+
+	closeOnce sync.Once
+	closed    chan struct{}
 }
 
 var _ executor.Executor = (*Executor)(nil)
@@ -70,6 +74,8 @@ func NewExecutor(
 
 		chainID:          chainID,
 		maxTableRowCount: maxTableRowCount,
+
+		closed: make(chan struct{}),
 	}
 	tblp.chBlockScope <- struct{}{}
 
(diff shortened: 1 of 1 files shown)


pkg/eventprocessor/impl/executor/impl/executor.go
```diff
@@ -80,6 +86,8 @@ func NewExecutor(
 func (ex *Executor) NewBlockScope(ctx context.Context, newBlockNum int64) (executor.BlockScope, error) {
 	select {
 	case <-ex.chBlockScope:
+	case <-ex.closed:
+		return nil, fmt.Errorf("executor is closed")
```
jsign, line 90: If `ex.closed` is closed, then we simply return an error instead of panicking.

The panicking makes sense to detect if other client is trying to call `NewBlockScope` concurrently, which is definitely wrong and a bug.

But notice how `Close()` works. It `<-ex.chBlockScope` to be sure there's no ongoing execution of a block, so it closes when no block is being executed. But... that means that a concurrent `NewBlockScope` could also happen at the same time, making `case <-ex.chBlockScope:` in L88 fail, thus defaulting to panic which isn't entirely correct. (There's no serial execution, but just being unlucky that we're closing).

This was fixed by an extra `ex.close` channel that is first closed by `Closed()`, and then it does L156. This way, before doing L156, we know that any concurrent `NewBlockScope` will error (instead of panicking) which is way better for graceful closing.

LMK if this sounds confusing!

pkg/eventprocessor/impl/executor/impl/executor.go
```diff
@@ -138,10 +146,17 @@ func (ex *Executor) getLastExecutedBlockNumber(ctx context.Context, txn *sql.Tx)
...
 	select {
 	case <-ctx.Done():
+		if err := ex.db.Close(); err != nil {
+			ex.log.Error().Err(err).Msg("forced close of database connection")
+		}
 		return errors.New("closing ctx done")
 	case <-ex.chBlockScope:
+		if err := ex.db.Close(); err != nil {
+			ex.log.Error().Err(err).Msg("closing database connection")
+		}
```
jsign, line 159: The db wasn't closed here.

pkg/eventprocessor/impl/executor/impl/executor.go
```diff
@@ -138,10 +146,17 @@ func (ex *Executor) getLastExecutedBlockNumber(ctx context.Context, txn *sql.Tx)
 // Close closes the processor gracefully. It will wait for any pending
 // batch to be closed, or until ctx is canceled.
 func (ex *Executor) Close(ctx context.Context) error {
+	ex.closeOnce.Do(func() { close(ex.closed) })
```
jsign, line 149: See comment above.

brunocalza (approved):
Looks correct to me 👍

PR #257, opened 2022-09-05
Update golangci-lint version, revive config, fixes

asutula:
Updated so `make lint` uses the current `golangci-lint` version `1.49.0`. 

With `golangci-lint` version `1.49.0`, and now using Go 0.19, the `revive` linter is quite powerful. I configured `golangci-lint` to use the recommended revive settings listed at https://github.com/mgechev/revive#recommended-configuration, but disabled the `package-comments` rule because I don't feel like I can provide meaningful package comments in one sweep right now. If we want to start using package comments, I'd say we should put the proper effort in where/when necessary, then we can enable the rule.

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -196,7 +196,12 @@ func (c *Client) List(ctx context.Context) ([]TableInfo, error) {
 		c.chain.ID,
 		c.wallet.Address().Hex(),
 	)
-	res, err := c.tblHTTP.Get(url)
+	req, err := http.NewRequest("GET", url, nil)
+	if err != nil {
+		return nil, fmt.Errorf("creating request: %v", err)
+	}
+	req = req.WithContext(ctx)
+	res, err := c.tblHTTP.Do(req)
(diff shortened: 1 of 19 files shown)


pkg/client/client.go
```diff
@@ -196,7 +196,12 @@ func (c *Client) List(ctx context.Context) ([]TableInfo, error) {
 		c.chain.ID,
 		c.wallet.Address().Hex(),
 	)
-	res, err := c.tblHTTP.Get(url)
+	req, err := http.NewRequest("GET", url, nil)
+	if err != nil {
+		return nil, fmt.Errorf("creating request: %v", err)
+	}
+	req = req.WithContext(ctx)
+	res, err := c.tblHTTP.Do(req)
```
asutula, line 204: This was a nice outcome of `revive` flagging that `ctx` wasn't being used. I figured I really should use it.

brunocalza (approved):
💯  really nice improvement

PR #258, opened 2022-09-08
Output formatting via new params

asutula:
This PR implements the new formatting parameters and behavior that we decided on. It does this in both the user controller (HTTP API) and JSON RPC API. Also updates the Go client `Read` func to allow use of new params.

The proposal for the new parameters included a `JSONStrings bool` parameter that would control whether or not nested JSON values are returned as JSON (the default and current behavior) or stringified JSON. This PR does NOT address that because I found that there is some important thinking and decision making to be done. We should discuss it more before implementing. TL;DR is that currently the logic of stringified JSON or not happens in the data store layer, not the formatting layer. This is the most optimal place to do it, but it doesn't feel right that way since it feels like it should be a formatting concern. But doing it as a formatting concern is less optimal performance wise. Anyway, I'll create an issue and we can discuss more.


.github/workflows/review.yml
```diff
@@ -17,7 +17,7 @@ jobs:
       - name: Install Go
         uses: actions/setup-go@v2
         with:
-          go-version: v1.19
+          go-version: 1.19
```
asutula, line 20: Noticed that our Go version names were inconsistent about using the `v`, and it caused me some confusion when managing required actions for merging PR, so just cleaned that up.

Makefile
```diff
@@ -34,6 +34,14 @@ system-sql-assets:
...
 .PHONY: system-sql-assets
 
+mocks: clean-mocks
+	go run github.com/vektra/mockery/v2@v2.14.0 --name='\b(?:SQLRunner|Tableland)\b' --recursive --with-expecter
+.PHONY: mocks
+
+clean-mocks:
+	rm -rf mocks
+.PHONY: clean-mocks
+
```
asutula, line 44: Got tired of dealing with manually created mocks, so introduced this tool for generating mocks. Using them in a couple places now, feels great.

internal/formatter/formatter.go
```diff
@@ -0,0 +1,149 @@
...
+// from the single property of the surrounding JSON object.
+// Default is false.
+func WithExtract(extract bool) FormatOption {
+	return func(fc *FormatConfig) {
+		fc.Extract = extract
+	}
+}
+
+// Format transforms the user rows according to the provided configuration, retuning raw json or jsonl bytes.
+func Format(userRows *tableland.TableData, opts ...FormatOption) ([]byte, FormatConfig, error) {
```
asutula, line 67: This function is used in both the user controller and json rpc service. The idea is that those services read data from the core api, `tableland.Tableland` which always returns `tableland.TableData` (formally called `tableland.UserData`), then pass that table data into the formatter with the appropriate `FormatOption`s derived from user input parameters. This function returns `[]byte` that are ready to return to the caller.

internal/formatter/formatter_test.go
```diff
@@ -0,0 +1,132 @@
...
+		{
+			name: "objects, extract, unwrap",
+			args: args{userRows: inputExtractable, output: Objects, extract: true, unwrap: true},
+			want: "\"bob\"\n\"jane\"",
+		},
+		{
+			name: "objects, extract, unwrap nested json",
+			args: args{userRows: inputExtractable2, output: Objects, extract: true, unwrap: true},
+			want: "{\"city\":\"dallas\"}\n{\"city\":\"dallas\"}",
+		},
```
asutula, line 97: Covering lots of cases here, feeling confident that formatting works as expected.

internal/router/controllers/user.go
```diff
@@ -183,148 +179,74 @@ func (c *UserController) GetTableRow(rw http.ResponseWriter, r *http.Request) {
...
+			msg := fmt.Sprintf("Error formatting data: %v", err)
+			_ = json.NewEncoder(rw).Encode(errors.ServiceError{Message: msg})
+			log.Ctx(r.Context()).Error().Err(err).Msg(msg)
+			return
+		}
+		if config.Unwrap && len(res.Rows) > 1 {
+			rw.Header().Set("Content-Type", "application/jsonl+json")
+		}
+		rw.WriteHeader(http.StatusOK)
+		_, _ = rw.Write(formatted)
```
asutula, line 205: Ok so this is handling the `GetTableRow` requests when `erc721` format is NOT requested. So we want to use our general formatting support here. `formatterOptions(r)` returns a slice of formatter options derived from the request query parameters. The call to `Format()` returns the formatted data along with the final configuration that was used to format the data. This config is essentially a composite of the default values with the user provided options. It can be used to make subsequent decisions based on the data formatting. On line 201, we make one of these decisions if the data is "unwrapped" and contains more than one row, we set the content type for `jsonl`.

internal/router/controllers/user.go
```diff
@@ -183,148 +179,74 @@ func (c *UserController) GetTableRow(rw http.ResponseWriter, r *http.Request) {
...
+		return
+	}
+
+	rw.WriteHeader(http.StatusOK)
 
-		_ = encoder.Encode(errors.ServiceError{Message: err.Error()})
+	if config.Unwrap && len(res.Rows) > 1 {
+		rw.Header().Set("Content-Type", "application/jsonl+json")
 	}
+	_, _ = rw.Write(formatted)
```
asutula, line 242: This is the handling for the general `/query` endpoint. Same logic as described above applies here.

internal/router/controllers/user.go
```diff
@@ -359,5 +269,69 @@ func (c *UserController) runReadRequest(
...
+	mode := r.URL.Query().Get("mode")
+	if mode == "list" {
+		v := true
+		c.unwrap = &v
+	} else if mode == "json" {
+		v := formatter.Objects
+		c.output = &v
+	}
+
+	return c, nil
```
asutula, line 336: This is just a helper to convert untypes request parameters into something typed. Note the support for legacy `mode` param. We can keep this around for a while in case some smart contracts or devs are using it.

asutula: I just chose some reasonable legacy things to support because they easily map to the new flags/behavior. Things like `csv` mode are dropped.

internal/router/controllers/user_test.go
```diff
@@ -65,17 +71,40 @@ func TestUserControllerInvalidColumn(t *testing.T) {
...
+				{Name: "personality"},
+				{Name: "aqua_power"},
+				{Name: "stamina_increase"},
+				{Name: "generation"},
+			},
+			Rows: [][]*tableland.ColumnValue{},
+		},
+		nil,
+	)
+
```
asutula, line 103: Switched over in this file to using the generated mocks.

internal/router/rpcservice/rpcservice.go
```diff
@@ -25,7 +27,10 @@ type RelayWriteQueryResponse struct {
 
 // RunReadQueryRequest is a user RunReadQuery request.
 type RunReadQueryRequest struct {
-	Statement string `json:"statement"`
+	Statement string  `json:"statement"`
+	Output    *string `json:"output"`
+	Unwrap    *bool   `json:"unwrap"`
+	Extract   *bool   `json:"extract"`
```
asutula, line 33: Updated RPC run read query request that includes the new formatting params.

internal/router/rpcservice/rpcservice.go
```diff
@@ -168,7 +173,31 @@ func (rs *RPCService) RunReadQuery(
...
+		opts = append(opts, formatter.WithExtract(*req.Extract))
+	}
+	if req.Unwrap != nil {
+		opts = append(opts, formatter.WithUnwrap(*req.Unwrap))
+	}
+
+	formatted, config, err := formatter.Format(res, opts...)
+	if err != nil {
+		return RunReadQueryResponse{}, fmt.Errorf("formatting result: %v", err)
+	}
```
asutula, line 194: Format data based on request params.

internal/router/rpcservice/rpcservice.go
```diff
@@ -168,7 +173,31 @@ func (rs *RPCService) RunReadQuery(
...
+
+	if config.Unwrap && len(res.Rows) > 1 {
+		return RunReadQueryResponse{}, errors.New("unwrapped results with more than one row aren't supported in JSON RPC API")
+	}
```
asutula, line 198: Important decision here. In the case of would-be `jsonl` data ("unwrapped" data) with more than one row, I chose not to support it in the JSON RPC API. This is because it would be invalid JSON and the JSON RPC response must be valid JSON. If there is just one row, the unwrapped data is just a single JSON object and is valid JSON. The other alternative would be to base64 encode the unwrapped data and then many rows could be included. This just seemed like the more simple thing to start with. We can always do the base64 encoding later, or it sounds like we'll probably just delete the RPC service anyway.

internal/router/rpcservice/rpcservice_test.go
```diff
@@ -0,0 +1,161 @@
...
+
+	"github.com/ethereum/go-ethereum/rpc"
+	"github.com/gorilla/mux"
+	"github.com/stretchr/testify/mock"
+	"github.com/stretchr/testify/require"
+	"github.com/textileio/go-tableland/internal/tableland"
+	"github.com/textileio/go-tableland/mocks"
+)
+
+func TestRunReadQueryManyRows(t *testing.T) {
```
asutula, line 17: Wanted to test `RunReadQuery` via JSON RPC calls just to be sure the wires were connecting properly in the RPC service and that the formatting and surrounding logic is working as expected.

internal/tableland/tableland.go
```diff
@@ -25,7 +91,7 @@ type TxnReceipt struct {
 
 // Tableland defines the interface of Tableland.
 type Tableland interface {
-	RunReadQuery(ctx context.Context, stmt string) (interface{}, error)
+	RunReadQuery(ctx context.Context, stmt string) (*TableData, error)
```
asutula, line 94: The core api always returns `*TableData` (previously `sqlstore.UserData`) so just making that explicit in the interface now. Also moved the definition of `TableData` to this file (it was in the store before) since it's a part of the core api now.

pkg/sqlstore/impl/user/rowstotabledata.go
```diff
@@ -4,10 +4,10 @@ import (
 	"database/sql"
 	"fmt"
 
-	"github.com/textileio/go-tableland/pkg/sqlstore"
+	"github.com/textileio/go-tableland/internal/tableland"
 )
 
-func rowsToJSON(rows *sql.Rows) (*sqlstore.UserRows, error) {
+func rowsToTableData(rows *sql.Rows) (*tableland.TableData, error) {
```
asutula, line 10: Just better naming in here.

pkg/sqlstore/store_user.go
```diff
@@ -2,79 +2,13 @@ package sqlstore
...
-
-// OtherUserValue creates a UserValue with the provided other value.
-func OtherUserValue(v interface{}) *ColValue {
-	return &ColValue{otherValue: v}
-}
-
 // UserStore defines the methods for interacting with user data.
 type UserStore interface {
-	Read(context.Context, parsing.ReadStmt) (interface{}, error)
+	Read(context.Context, parsing.ReadStmt) (*tableland.TableData, error)
```
asutula, line 12: `Read` only returns `*tableland.TableData`, so being explicit about it.

asutula:
cc @carsonfarmer this PR is going to be relevant to the JS SDK and CLI.

pkg/client/client.go
```diff
@@ -279,22 +279,58 @@ func (c *Client) Create(ctx context.Context, schema string, opts ...CreateOption
...
+	return func(rrqr *rpcservice.RunReadQueryRequest) {
+		v := true
+		rrqr.Unwrap = &v
 	}
+}
 
-	// TODO: Make this do something better than returning a json string.
-	return string(b), nil
+// Read runs a read query with the provided opts and unmarshals the results into target.
+func (c *Client) Read(ctx context.Context, query string, target interface{}, opts ...ReadOption) error {
```
asutula, line 322: Updated client API for reading and unmarshaling data into user provided value.

internal/formatter/formatter.go
```diff
@@ -0,0 +1,148 @@
...
+	}
+	return objects
+}
+
+func extract(in []interface{}) ([]interface{}, error) {
+	extracted := make([]interface{}, len(in))
+	for i, item := range in {
+		object := item.(map[string]interface{})
+		if len(object) != 1 {
+			return nil, fmt.Errorf("can only extract values for result sets with one column but this has %d", len(object))
```
joewagner, line 125: @asutula I setup the SDK to allow the extract option and some end to end tests.  I might be missing something, but whether the query returns 1, 2, or 3 rows I'm always getting an error with:
`formatting result: extracting values: can only extract values for result sets with one column but this has 2`
The SDK branch is here https://github.com/tablelandnetwork/js-tableland/tree/joe/output-formatting
and the tests are here https://github.com/tablelandnetwork/local-tableland/tree/joe/npx-test

joewagner: Disregard, I was confused on what the extract options does.

internal/router/rpcservice/rpcservice.go
```diff
@@ -168,7 +173,31 @@ func (rs *RPCService) RunReadQuery(
 	if err != nil {
 		return RunReadQueryResponse{}, fmt.Errorf("calling RunReadQuery: %v", err)
 	}
-	return RunReadQueryResponse{Result: res}, nil
+	var opts []formatter.FormatOption
```
brunocalza, line 176: Cleaner code suggestion: Create a method `FormatOpts()` on `RunReadQueryRequest`, and put the logic of lines L176-189 in there. So, the code here would just be `opts, err := req.FormatOpts()`

internal/router/rpcservice/rpcservice.go
brunocalza: Makes sense. Would make sense to put this inside the `Format` and maybe eliminating the need to return the `config` value?

internal/router/rpcservice/rpcservice_test.go
brunocalza: Nice! The mock library is very handy

internal/router/rpcservice/rpcservice.go
brunocalza: The above would imply an extra error to make different returns according to the error.

I just saw the `user.go` file and it has the same logic.

brunocalza (approved):
Looks very good to me 💯 Left a couple of comments for your consideration :)

joewagner:
I updated the open api spec to match these changes [here](https://github.com/tablelandnetwork/go-tableland/commit/d31e408c8483dab53006d2178635365fabbfb074)

internal/router/rpcservice/rpcservice.go
asutula: Nice, I like that idea. Doing it.

internal/router/rpcservice/rpcservice.go
asutula: Hmm. The issue is that this logic equates to an error only in the context of the RPC service. In `user.go`, it's not considered an error, only that the content type header is set differently. Or maybe I'm not quite following what you're suggesting?

asutula: Yea, see `user.go` it is:
```
		 if config.Unwrap && len(res.Rows) > 1 {
			rw.Header().Set("Content-Type", "application/jsonl+json")
		}
```

brunocalza: Oh, I see. It's an error for RPC but not for HTTP. All good then.

PR #264, opened 2022-09-16
Implements state hash calculation every X block

brunocalza:
# Summary
This PR implements a way of calculating the hash of a particular state of the database for every X block

# Context
As we start to onboard external validators, we must ensure we have a healthy network of validators. Having a healthy network of validators begins by making sure validators are materializing precisely the same database state. If that's not true, Tableland users are not able to trust our network.

So, this PR intends to collect a state hash metric for every X block. That metric is later published to a central telemetry server (yet to be built), and from there we can check if validators are materializing events correctly and agree on the same state for a given block.

# Implementation overview

## What's the right moment during execution to calculate the state hash?

The database is constantly changing as events come in, and each chain updates the database concurrently. So choosing a specific moment to get the database state to calculate the hash is not trivial.  We have to pick a moment where we know all validators should have the same state and any other changes cannot happen concurrently. Some thought processes that shaped our implementation:

- Since each chain processes events in its own _goroutine_, we can't calculate the global state of the database. It's not like we can freeze all _goroutines_ to calculate the database state. That means the state calculation has to be per chain.
- Inside each chain execution, blocks of events are processed serially. That means we can pick any moment between two block execution, and at that moment we know the state of the database will be frozen for all data of that chain. That's the right moment to do the calculation.

## How is the hash calculated?

The general idea is to fetch the name and the schema of all tables from a chain and all system tables, such as `registry`, `system_acl`, `system_txn_receipts` & `system_controller`. Then for each table, fetch all rows and columns. And then pass all that data to a hasher. So, we would be passing the table name, its schema, and all its data to a hasher. 

We have to make sure that the state we grab from the database is deterministic. So we have to exclude nondeterministic columns from system tables such as `created_at` and `updated_at`. For user tables, they all should have deterministic data, according to our SQL spec. No need to worry about it. And also, we have to use an `ORDER BY` clause to fetch all the data in the same order. 

## What happens after the hash is calculated?

The hash is collected and stored in a new local database called `metrics.db`. Later, another process will publish this metric to a central server. 

You can start the review by going through the bullet points. That will give a good overview of the implementation. Specifics will be commented on the `Files changed` tab.

- The hash is calculated and collected inside the event processor. Check lines [213-218](https://github.com/tablelandnetwork/go-tableland/pull/264/files#diff-046a77aeba18c13c7f18f3fdbaa1451a351c72c0ce2ee9de4d0f1a0d46193d83R213-R218)
    - Check the [`calculateHash` implementation](https://github.com/tablelandnetwork/go-tableland/pull/264/files#diff-046a77aeba18c13c7f18f3fdbaa1451a351c72c0ce2ee9de4d0f1a0d46193d83R288-R307). 
    - It makes use of a new method [StateHash](https://github.com/tablelandnetwork/go-tableland/pull/264/files#diff-6d26a4b1c786f7e0c5669cae8ec7e7af0206df2851916c9665dcedf2b167d772R168) in the `blockScope` object.
- A new package [dbhash](https://github.com/tablelandnetwork/go-tableland/pull/264/files#diff-068c52bbc04999650cf504f883632ba19447b12aee2a71e7917842d380daf2f6) was created to isolate the logic of calculating the hash. It's a fairly generic implementation that allows you to pass the queries you want to grab the state. So, the [StateHash](https://github.com/tablelandnetwork/go-tableland/pull/264/files#diff-6d26a4b1c786f7e0c5669cae8ec7e7af0206df2851916c9665dcedf2b167d772R168) defines the query it needs and delegate the hash calculation to [dbhash.DatabaseStateHash](https://github.com/tablelandnetwork/go-tableland/pull/264/files#diff-068c52bbc04999650cf504f883632ba19447b12aee2a71e7917842d380daf2f6R13)
- A new package [telemetry](https://github.com/tablelandnetwork/go-tableland/pull/264/files#diff-1daf3b94af9e46763e64ff02c31240f3c3b1421994d4d56e05769dfd4f765ba4) was created that acts as a library that allows collecting metrics such as [StateHashMetric](https://github.com/tablelandnetwork/go-tableland/pull/264/files#diff-4ab1048c1ff4b2de22e2a53e11ce2f4dd2e9c4e18949ee63541ee5854c8778b4R43), and persisting them locally.
- Metrics are persisted in another database called `metrics.db`.


cmd/api/main.go
```diff
@@ -96,6 +98,18 @@ func main() {
...
+		"file://%s?_busy_timeout=5000&_foreign_keys=on&_journal_mode=WAL",
+		path.Join(dirPath, "metrics.db"),
+	)
+
+	metricsStore, err := storage.New(metricsDatabaseURL)
+	if err != nil {
+		log.Fatal().Err(err).Msg("creating metrics store")
+	}
+
+	telemetry.SetMetricStore(metricsStore)
```
brunocalza, line 111: Sets the store of the `telemetry` library

pkg/dbhash/dbhash.go
```diff
@@ -0,0 +1,120 @@
...
+	for rows.Next() {
+		var name, stmt string
+		if err = rows.Scan(
+			&name,
+			&stmt,
+		); err != nil {
+			return fmt.Errorf("schema scan row: %s", err)
+		}
+
+		_, _ = writer.Write([]byte(name))
```
brunocalza, line 48: We write both table name and its schema to the hasher

pkg/dbhash/dbhash.go
```diff
@@ -0,0 +1,120 @@
...
+
+	for rows.Next() {
+		var name, stmt string
+		if err = rows.Scan(
+			&name,
+			&stmt,
+		); err != nil {
+			return fmt.Errorf("schema scan row: %s", err)
+		}
+
```
brunocalza, line 47: we write both table's name and its schema to the hasher

pkg/dbhash/dbhash.go
```diff
@@ -0,0 +1,120 @@
...
+			scanCallArgs[i] = &rawBuffer[i]
+		}
+
+		for tableRows.Next() {
+			if err := tableRows.Scan(scanCallArgs...); err != nil {
+				return fmt.Errorf("table row scan: %s", err)
+			}
+
+			for _, col := range rawBuffer {
+				_, _ = writer.Write(col)
```
brunocalza, line 79: we write the raw byte for a specific row and column to the hasher

pkg/dbhash/dbhash.go
```diff
@@ -0,0 +1,120 @@
...
+	if err := databaseStateWriter(ctx, tx, h, config); err != nil {
+		return "", fmt.Errorf("database hash writer failed: %s", err)
+	}
+
+	return hex.EncodeToString(h.Sum(nil)), nil
+}
+
+func databaseStateWriter(ctx context.Context, tx *sql.Tx, writer io.Writer, c *Config) error {
+	// get all tables from db and associated schema
+	rows, err := tx.QueryContext(ctx, c.FetchSchemasQuery)
```
brunocalza, line 31: The query to fetch all tables is configurable. But there's a caveat: It kind of expects the query to have the columns (`tbl_name` and `sql`) that come from `sqlite_schema` table to the whole thing to work properly. There's not a validation for that, it kind of expects the client to do the right thing.

pkg/dbhash/dbhash.go
```diff
@@ -0,0 +1,120 @@
...
+
+	h := sha1.New()
+	if err := databaseStateWriter(ctx, tx, h, config); err != nil {
+		return "", fmt.Errorf("database hash writer failed: %s", err)
+	}
+
+	return hex.EncodeToString(h.Sum(nil)), nil
+}
+
+func databaseStateWriter(ctx context.Context, tx *sql.Tx, writer io.Writer, c *Config) error {
```
brunocalza, line 29: This is the main logic of gathering the database state through an `io.Writer`. It's quite straightforward. It fetches all tables by executing the query configured at `FetchSchemasQuery`. For each table, it executes the query generated by `PerTableQueryFn`. See [L51](https://github.com/tablelandnetwork/go-tableland/pull/264/files#diff-068c52bbc04999650cf504f883632ba19447b12aee2a71e7917842d380daf2f6R51). A long the way it sends the bytes to the `io.Writer`, avoiding extra loops.

pkg/dbhash/dbhash.go
```diff
@@ -0,0 +1,120 @@
...
+	"context"
+	"crypto/sha1"
+	"database/sql"
+	"encoding/hex"
+	"fmt"
+	"io"
+)
+
+// DatabaseStateHash calculates the hash of some state of the database according to the options passed.
+func DatabaseStateHash(ctx context.Context, tx *sql.Tx, opts ...Option) (string, error) {
```
brunocalza, line 13: This is the method that the package exports to the external word making it possible to calculate the hash of some state of the database. I made it generic in the sense that you can pass the query for the data you want to calculate the state on. This decision was mostly so I could come up with some simple tests to make sure the logic of gathering the bytes and hashing was correct, not because I wanted to make a future-proof hashing solution (although i may help in that case as well)

pkg/telemetry/telemetry.go
```diff
@@ -0,0 +1,84 @@
...
+// Can only be called once, and should be called before Collect.
+func SetMetricStore(s MetricStore) {
+	once.Do(func() {
+		metricStore = s
+	})
+}
+
+// Collect collects the metric by persisting locally for later publication.
+// If Collect is called before setting the metric store, it will simply log the metric without persisting it.
+func Collect(ctx context.Context, metric interface{}) error {
```
brunocalza, line 44: I opted for a global function on this library to make it really easy to collect a metric in a non-invasive manner. You can simply call `telemetry.Collect`. No need to inject any dependency on clients. Of course that comes with some trade-off. We have the [global variables](https://github.com/tablelandnetwork/go-tableland/pull/264/files#diff-1daf3b94af9e46763e64ff02c31240f3c3b1421994d4d56e05769dfd4f765ba4R13-R20) that mess with concurrency, and make the code error prone.

pkg/telemetry/telemetry.go
```diff
@@ -0,0 +1,84 @@
...
+	mu.Lock()
+	defer mu.Unlock()
+	if metricStore == nil {
```
brunocalza, line 47: I opted to not return an error with clients who forgot to call `SetMetricStore`. That means that client code will not get interrupted because of a metric that has no store set. The only thing is that the metric will not be persisted. That is good, so I don't have to call `SetMetricStore` for all existing tests.

pkg/telemetry/telemetry.go
```diff
@@ -0,0 +1,84 @@
...
+		return nil
+	}
+
+	if c == nil {
+		c = &collector{
+			s: metricStore,
+		}
+	}
+
+	switch v := metric.(type) {
```
brunocalza, line 58: This is another decision I made. A metric to satisfy a specific set of interfaces. In this case, we only have only: `StateHash`.  I could have let the door open, and any structure could be considered a metric. However, I thought letting the door open could lead to a case where we have too much garbage in the metrics database. With this restriction, we can be more thoughtful about which kind of data we add to the metrics database.

cmd/api/main.go
```diff
@@ -96,6 +98,18 @@ func main() {
 		log.Fatal().Err(err).Msg("creating user store")
 	}
 
+	metricsDatabaseURL := fmt.Sprintf(
```
brunocalza, line 101: Decided to go with another database called `metrics.db` to keep things decoupled. Metrics are not part of the protocol and are still an experimental thing.

pkg/eventprocessor/impl/executor/impl/blockscope.go
```diff
@@ -162,6 +165,54 @@ func (bs *blockScope) TxnReceiptExists(ctx context.Context, txnHash common.Hash)
 	return true, nil
 }
 
+func (bs *blockScope) StateHash(ctx context.Context, chainID tableland.ChainID) (executor.StateHash, error) {
+	hash, err := dbhash.DatabaseStateHash(ctx, bs.txn, []dbhash.Option{
+		dbhash.WithFetchSchemasQuery(
```
brunocalza, line 170: This query fetches all tables for a particular chain and all system tables that are relevant to state hash calculation

pkg/eventprocessor/impl/executor/impl/blockscope.go
```diff
@@ -162,6 +165,54 @@ func (bs *blockScope) TxnReceiptExists(ctx context.Context, txnHash common.Hash)
...
+			    WHERE name NOT LIKE 'sqlite_%%'  
+				AND name LIKE '%%\_%d\_%%' ESCAPE '\'
+				AND type = 'table'
+				UNION ALL
+				SELECT tbl_name, sql 
+				FROM sqlite_schema
+				WHERE name in ('registry', 'system_acl', 'system_controller', 'system_txn_receipts')
+				ORDER BY tbl_name COLLATE nocase;`, chainID),
+		),
+		dbhash.WithPerTableQueryFn(func(tableName string) string {
```
brunocalza, line 182: This is a bit ugly but each system table requires a particular query to fetch its data. That's because we have to exclude non-deterministic columns such as `created_at` and add specific `ORDER BY` clauses.

pkg/eventprocessor/impl/executor/impl/blockscope.go
```diff
@@ -162,6 +165,54 @@ func (bs *blockScope) TxnReceiptExists(ctx context.Context, txnHash common.Hash)
...
+							FROM system_controller 
+							WHERE chain_id = %d
+							ORDER BY table_id`, chainID)
+			case "system_txn_receipts":
+				return fmt.Sprintf(`SELECT chain_id, block_number, index_in_block, txn_hash, error, table_id 
+							FROM system_txn_receipts 
+							WHERE chain_id = %d 
+							ORDER BY table_id, block_number, index_in_block`, chainID)
+			default:
+				return fmt.Sprintf("SELECT * FROM %s", tableName)
```
brunocalza, line 205: I guess this is deterministic for all user tables. (maybe adding an `ORDER BY rowid` to be sure.)

pkg/eventprocessor/eventprocessor.go
```diff
@@ -12,13 +12,15 @@ import (
 type Config struct {
 	BlockFailedExecutionBackoff time.Duration
 	DedupExecutedTxns           bool
+	HashCalcStep                int64
```
brunocalza, line 15: A new config is used to slow the pace of hash calculation

pkg/eventprocessor/eventprocessor.go
```diff
@@ -12,13 +12,15 @@ import (
...
 }
 
 // DefaultConfig returns the default configuration.
 func DefaultConfig() *Config {
 	return &Config{
 		BlockFailedExecutionBackoff: time.Second * 10,
 		DedupExecutedTxns:           false,
+		HashCalcStep:                100,
```
brunocalza, line 23: I did not add any entries to `config.json`. Pretty much relying on the default. What would be a good default? 100 feels too low. I put it 100 because it's good to see it on `staging` without waiting too much.

pkg/eventprocessor/impl/eventprocessor.go
```diff
@@ -139,6 +142,7 @@ func (ep *EventProcessor) startDaemon() error {
 		return fmt.Errorf("get last executed block number: %s", err)
 	}
 	ep.mLastProcessedHeight.Store(fromHeight)
+	ep.nextHashCalcBlockNumber = nextMultipleOf(fromHeight, ep.config.HashCalcStep)
```
brunocalza, line 145: When we start `EventProcessor` we have to figure out the next block which it will do the hash calculation. It is the next block greater than `fromHeight` that is multiple of  `ep.config.HashCalcStep`.

jsign (approved):
Looks good to me! 
Many comments, but there're some particular ones that I think might worth some changes.

It'll be interesting to see how this plays out when we start collecting and comparing data. If two nodes differ in the hash, we might be in a "black box" situation on how to debug it. 

I think the main challenge will be that the actual database that calculated the hash would have changed whenever we detect a mismatch of hashes, so we can't debug it. We can think about this soon, but one option could be (with an optional flag) to create a backup just before the hash calc, so at least we can later debug. Or.. backup triggering is always tied to hash calculation, which sounds nice and kills two birds with one bullet. (Not work for this PR, just sharing thoughts)

go.mod
```diff
@@ -6,6 +6,7 @@ require (
 	cloud.google.com/go/logging v1.5.0
 	github.com/XSAM/otelsql v0.16.0
 	github.com/ethereum/go-ethereum v1.10.23
+	github.com/golang-migrate/migrate v3.5.4+incompatible
```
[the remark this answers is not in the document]
brunocalza: thanks! `go mod tidy` did the trick

pkg/eventprocessor/eventprocessor.go
brunocalza: Do you think we should add entries `config.json`? It seems we'd like to avoid validators messing with that config.

pkg/eventprocessor/impl/executor/impl/blockscope.go
```diff
@@ -10,17 +10,19 @@ import (
...
-	log    zerolog.Logger
-	parser parsing.SQLValidator
-	acl    tableland.ACL
+	txn         *sql.Tx
+	log         zerolog.Logger
+	parser      parsing.SQLValidator
+	acl         tableland.ACL
+	blockNumber int64
 
 	scopeVars scopeVars
```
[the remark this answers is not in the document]
brunocalza: Makes sense

pkg/eventprocessor/eventprocessor.go
jsign: The thing is that 100 blocks for Optimsim is almost "nothing" since it produces blocks at a high rate ([see here](https://optimistic.etherscan.io/txs)). 

So that would mean it will probably be doing hashing in all events. But for example in Ethereum we'd get one block per 12sec.

I'm not sure we can avoid having a knob in the config.

pkg/eventprocessor/impl/executor/impl/blockscope.go
```diff
@@ -162,6 +165,54 @@ func (bs *blockScope) TxnReceiptExists(ctx context.Context, txnHash common.Hash)
 	return true, nil
 }
 
+func (bs *blockScope) StateHash(ctx context.Context, chainID tableland.ChainID) (executor.StateHash, error) {
+	hash, err := dbhash.DatabaseStateHash(ctx, bs.txn, []dbhash.Option{
+		dbhash.WithFetchSchemasQuery(
+			fmt.Sprintf(`SELECT tbl_name, sql 
+				FROM sqlite_schema
+			    WHERE name NOT LIKE 'sqlite_%%'
```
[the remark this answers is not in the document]
brunocalza: We should not have any `sqlite_*` tables. 

It is more a safe check trying to be future-proof. Example: `sqlite_sequence`is created when AUTOINCREMENT is used. There are other cases such as `sqlite_stat` or `sqlite_user`.

pkg/eventprocessor/impl/executor/impl/blockscope.go
```diff
@@ -162,6 +165,54 @@ func (bs *blockScope) TxnReceiptExists(ctx context.Context, txnHash common.Hash)
 	return true, nil
 }
 
+func (bs *blockScope) StateHash(ctx context.Context, chainID tableland.ChainID) (executor.StateHash, error) {
+	hash, err := dbhash.DatabaseStateHash(ctx, bs.txn, []dbhash.Option{
+		dbhash.WithFetchSchemasQuery(
+			fmt.Sprintf(`SELECT tbl_name, sql 
+				FROM sqlite_schema
+			    WHERE name NOT LIKE 'sqlite_%%'  
+				AND name LIKE '%%\_%d\_%%' ESCAPE '\'
```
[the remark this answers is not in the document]
brunocalza: I needed to escape `%` with another `%` for `Sprintf`

pkg/eventprocessor/impl/executor/impl/blockscope.go
```diff
@@ -162,6 +165,54 @@ func (bs *blockScope) TxnReceiptExists(ctx context.Context, txnHash common.Hash)
...
+				AND type = 'table'
+				UNION ALL
+				SELECT tbl_name, sql 
+				FROM sqlite_schema
+				WHERE name in ('registry', 'system_acl', 'system_controller', 'system_txn_receipts')
+				ORDER BY tbl_name COLLATE nocase;`, chainID),
```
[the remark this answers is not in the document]
brunocalza: Yeah. No need for that 👍

pkg/telemetry/metrics.go
```diff
@@ -0,0 +1,47 @@
...
+	BlockNumber() int64
+	Hash() string
+}
+
+// StateHashMetric defines a state hash metric.
+type StateHashMetric struct {
+	ChainID     int64  `json:"chain_id"`
+	BlockNumber int64  `json:"block_number"`
+	Hash        string `json:"hash"`
+}
```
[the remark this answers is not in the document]
brunocalza: Good call. Usually, it's a good practice to have two versions. One for the outer struct (in this case `Metric`) and the other for the `Payload` (in this case `StateHashMetric`). I'll add it 👍

pkg/telemetry/storage/db.go
```diff
@@ -0,0 +1,108 @@
...
+// StoreMetric persists a metric.
+func (db *TelemetryDatabase) StoreMetric(ctx context.Context, metric telemetry.Metric) error {
+	payloadJSON, err := metric.Serialize()
+	if err != nil {
+		return fmt.Errorf("marshal json: %s", err)
+	}
+
+	_, err = db.sqlDB.ExecContext(ctx,
+		`INSERT INTO system_metrics ("timestamp", "type", "payload", "published") VALUES (?1, ?2, ?3, ?4)`,
+		metric.Timestamp.UnixMilli(), metric.Type, payloadJSON, 0,
```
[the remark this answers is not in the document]
brunocalza: I thought `seconds` was not a good enough granularity for the timestamp of a metric. Is that thinking correct?

pkg/eventprocessor/impl/executor/impl/blockscope.go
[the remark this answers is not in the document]
jsign: Cool, interestingly `sqlite_schema` is always ignored apparently since that always exists.

pkg/telemetry/telemetry.go
```diff
@@ -0,0 +1,84 @@
...
+}
+
+// MetricStore specifies the methods for persisting a metric.
+type MetricStore interface {
+	StoreMetric(context.Context, Metric) error
+	Close() error
+}
+
+// SetMetricStore sets the store implementation.
+// Can only be called once, and should be called before Collect.
```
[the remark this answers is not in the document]
brunocalza: Good suggestion!

pkg/telemetry/storage/db.go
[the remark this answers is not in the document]
jsign: Yeah, forget what I said, that's fine. The thing is that we can't even trust this timestamp since comes from a place we don't know, but at least in relative terms milliseconds can give us relative order which is fine. Not sure we would need to distinguish ordering at the millisecond level for a metric but who knows.
Sounds fine.

pkg/telemetry/telemetry.go
```diff
@@ -0,0 +1,84 @@
...
+	if metricStore == nil {
+		log.Warn().Msg("no metric store was set")
+		return nil
+	}
+
+	if c == nil {
+		c = &collector{
+			s: metricStore,
+		}
+	}
```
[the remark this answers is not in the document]
brunocalza: The third option is great. Much better. No need for two global variables.

PR #269, opened 2022-09-20
eventfeed/eventspersistance: exclude incorrect error log

jsign:
Looking at the logs I see this error message `sql: transaction has already been committed or rolled back`, which looked odd.

That message corresponds to `sql.ErrTxDone`, which signals that you can't do any further operation in the transaction since it was already finished (Committed or Rollbacked).

Our idiom was fine since it is a good practice always to defer a rollback to cover any error-early-return case, so we roll back the transaction. If the execution was fine (and thus Committed), Rollback is a noop. 

The problem here was that it only makes sense to log the Rollback error if isn't `sql.ErrTxDone` (which signals that there was a commit since all went fine).

### DIFF
--- a/pkg/eventprocessor/eventfeed/impl/eventfeed.go
+++ b/pkg/eventprocessor/eventfeed/impl/eventfeed.go
@@ -2,6 +2,7 @@ package impl
 
 import (
 	"context"
+	"database/sql"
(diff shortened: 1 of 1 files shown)


brunocalza (approved):
good catch!

PR #272, opened 2022-09-21
Use extract for mode=list

asutula:
This is to provide compatibility with urls using `mode=list` for NFT metadata


sanderpick (approved):
can't say i understand the change cause i haven't dived into the logic, but considering it's a hotfix... lgtm!

PR #273, opened 2022-09-22
Change EventFeed chain sync mechanics to impact Ethereum node SaaS costs

jsign:
# Summary
This PR changes the mechanics of how `EventFeed` works. We detected new blocks using a subscription-api to be notified about new blocks in the chain. That's fine since it's an efficient way to know when to get recent events from chains.

It has a downside related to Ethereum node API calls, mostly in supported chains with fast block production, such as Optimism chains. In those cases, we're having a lot of activity in these subscriptions and other downstream API calls with reactivity that isn't required.

This PR switches the mechanics to do polling with a configurable delay which means:
1) The validator operator can put a cap on how fast to react getting new blocks to have an efficient tradeoff. For example, receiving a stream of new blocks of 10 blocks per second is inefficient since 99% of those blocks won't have events of interest. 
2) On chains that have a ~fixed block production rate, e.g. Ethereum ~12s, we can set a poll time of ~12s which usually can be hardly distinguishable from a subscription by the nature of block propagation (and if we're strict, potentially block-offset consideration for reorg coverage)
3) Apart from point 2) regarding that can be ~equivalent to subscription mode, the validator operator can stretch the poll time to sync every two or more blocks on average to drive the costs down, even more, making a tradeoff between reactivity and expenses.

# Context
This situation was bubbled by looking at our validator Ethereum node API SaaS costs. We expect this will help put a cap on costs in very active chains with a disproportionate share of the expenses in our validator.

# Implementation overview
The implementation doesn't have any hacks or complexities apart from what one would naturally expect, but here're some extra notes:
1) The previous `NewBlockTimeout` configuration attribute was deleted since this configuration was used for subscription mode. The value was used to configure how the subscription healthiness was managed. Now it isn't needed.
2) A new `NewBlockPollFreq` configuration attribute was created, indicating the poll frequency for the chain.

The `EventFeed` goroutines got simplified a bit due to the nature of the change.

I've also tried to reduce some noise in logging some instrumented calls that don't add much information and make the validator output very noisy. We should probably keep working on this to improve the high logs signal even at Debug level.

This PR has been running for a while in staging and the dashboard looks healthy reg sync metrics.

### DIFF
--- a/docker/deployed/testnet/api/config.json
+++ b/docker/deployed/testnet/api/config.json
@@ -31,17 +31,17 @@
     "Analytics": {
         "FetchExtraBlockInfo": true
     },
-    "Backup" : {
-		"Enabled": true, 
-		"Dir": "backups",               
-		"Frequency": 240,  
-		"EnableVacuum": true,      
-		"EnableCompression": true,
-        "Pruning" : {
+    "Backup": {
+        "Enabled": true,
+        "Dir": "backups",
+        "Frequency": 240,
+        "EnableVacuum": true,
+        "EnableCompression": true,
+        "Pruning": {
             "Enabled": true,
             "KeepFiles": 5
         }
-	},
+    },
     "Chains": [
         {
             "Name": "Optimism Kovan",
@@ -56,7 +56,7 @@
             },
             "EventFeed": {
                 "ChainAPIBackoff": "15s",
-                "NewBlockTimeout": "30s",
+                "NewBlockPollFreq": "5s",
                 "MinBlockDepth": 0,
                 "PersistEvents": true
             },
@@ -84,7 +84,7 @@
             },
             "EventFeed": {
                 "ChainAPIBackoff": "15s",
-                "NewBlockTimeout": "45s",
+                "NewBlockPollFreq": "10s",
                 "MinBlockDepth": 1,
                 "PersistEvents": true
             },
@@ -112,7 +112,7 @@
             },
             "EventFeed": {
                 "ChainAPIBackoff": "15s",
-                "NewBlockTimeout": "30s",
+                "NewBlockPollFreq": "5s",
                 "MinBlockDepth": 1,
                 "PersistEvents": true
             },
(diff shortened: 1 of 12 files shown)


docker/deployed/testnet/api/config.json
```diff
@@ -31,17 +31,17 @@
...
+        "Frequency": 240,
+        "EnableVacuum": true,
+        "EnableCompression": true,
+        "Pruning": {
             "Enabled": true,
             "KeepFiles": 5
         }
-	},
+    },
     "Chains": [
```
jsign, line 45: The rationale for selecting `NewBlockPollFreq` is still quite aggressive to have a similar UX for users as the previous subscription mode. We could hand tune these values a bit further to drive costs down even more, but probably we'll start making tradeoffs with _reactivity_.

For now, my idea is to select values to don't make tradeoffs about UX and see how much this can help with costs. After that, we can keep meditating it makes sense to be less reactive considering UX tradeoffs (if any).

The idea is that the values are the following:
- For fast chains that also don't have an expected block production since they're creating blocks one per transaction (e.g., Optimism), we use 5s.
- For other chains that have an expected average block time (e.g., Eth mainnet, Polygon Mumbai, etc.), we select that block time or slightly less. The value shouldn't be perfect since block time isn't guaranteed and dances around an expected value.

pkg/eventprocessor/eventfeed/impl/eventfeed.go
```diff
@@ -163,6 +163,7 @@ func (ef *EventFeed) Start(
 				// in the next head. Probably the API can have transient unavailability.
 				ef.log.Warn().Err(err).Msgf("filter logs from %d to %d", fromHeight, toHeight)
 				if strings.Contains(err.Error(), "read limit exceeded") ||
+					strings.Contains(err.Error(), "Log response size exceeded") ||
```
jsign, line 166: While testing the PR locally, I noticed that Arbitrum use a slightly different error message that we weren't detecting. 

I dug if this was happening in the cloud validator, and wasn't the case. I think this is because the cloud validator was always syncing this chain from new blocks, but in my case was lagged behind and thus syncing events in bigger block ranges which can trigger this error.

In any case, we detect it and do the usual shrinking of block ranges. This fixed the problem in my local node as expected.

brunocalza (approved):
Looks good! Much simpler indeed

PR #274, opened 2022-09-22
Adjust local config to add hash calculation step

brunocalza:
`local` was failing because the default 0 for `HashCalculationStep` config is not accepted. This PR adds a value to `local/api/config.json` to avoid 0 being the default. 

cc @joewagner


joewagner:
> cc @joewagner

Works great for me

jsign (approved):

PR #279, opened 2022-09-23
General use full stack test harness, migrate health bot to Go client

asutula:


cmd/api/main.go
```diff
@@ -115,15 +116,35 @@ func main() {
...
 	)
+	if err != nil {
+		log.Fatal().Err(err).Msg("creating system store")
+	}
+	systemService, err = systemimpl.NewInstrumentedSystemSQLStoreService(systemService)
+	if err != nil {
+		log.Fatal().Err(err).Msg("instrumenting system sql store")
+	}
+
+	router := router.ConfiguredRouter(mesaService, systemService, config.HTTP.MaxRequestPerInterval, rateLimInterval)
```
asutula, line 147: We now need to build these dependencies at a higher level to pass into `router.ConfiguredRouter` because of the API change there.

cmd/healthbot/config.go
```diff
@@ -22,17 +22,18 @@ type config struct {
...
+	Signer struct {
+		PrivateKey string `default:""`
+	}
 	Probe struct {
 		CheckInterval  string `default:"15s"`
 		ReceiptTimeout string `default:"20s"`
-		SIWE           string `default:""`
 		Tablename      string `default:""`
 	}
 }
```
asutula, line 39: I'm making some assumptions here about the possible structure of the json file that is read for the healthbot config. `docker/local/api/config.json` includes a `Signer.PrivateKey` property that I'm leveraging. I see that `docker/deployed/testnet/healthbot/config.json` doesn't have that property, but I'm hoping it would be ok to add it. There is no need for `Probe.SIWE` because the Go client creates the SIWE string internally by using the provided private key. Let's talk about this.

cmd/healthbot/counterprobe/counterprobe.go
```diff
@@ -4,17 +4,14 @@ import (
...
 	"errors"
 	"fmt"
-	"net/url"
 	"sync"
 	"time"
 
 	"github.com/rs/zerolog"
 	logger "github.com/rs/zerolog/log"
-	"github.com/textileio/go-tableland/internal/router/rpcservice"
+	"github.com/textileio/go-tableland/pkg/client"
```
asutula, line 12: All the changes is this file are to move healthbot from using raw json rpc calls to using the Go client.

cmd/healthbot/counterprobe/counterprobe.go
```diff
@@ -43,8 +40,7 @@ type CounterProbe struct {
 // New returns a *CounterProbe.
 func New(
 	chainName string,
-	endpoint string,
-	siwe string,
+	client *client.Client,
```
asutula, line 43: Passing in a fill `*client.Client` here, makes it nicer for tests and also just been a fan of dependency injection as much as possible (vs passing in configuration for dependencies to-be-created).

cmd/healthbot/counterprobe/counterprobe.go
```diff
@@ -144,57 +132,39 @@ func (cp *CounterProbe) healthCheck(ctx context.Context) (int64, error) {
...
+	if err := cp.client.Read(
+		ctx,
+		fmt.Sprintf("select counter from %s", cp.tableName),
+		&counter,
+		client.ReadExtract(),
+		client.ReadUnwrap(),
+	); err != nil {
+		return 0, fmt.Errorf("calling client read: %s", err)
+	}
+	return counter, nil
```
asutula, line 169: Much cleaner using the client and just unmarshaling directly into our `counter` value.

cmd/healthbot/counterprobe/counterprobe_test.go
```diff
@@ -2,19 +2,80 @@ package counterprobe
...
+	stack := testutils.CreateFullStack(t, testutils.Deps{Tableland: tbl})
+
+	cp, err := New("optimism", stack.Client, "Runbook_24", time.Second, time.Second*10)
+	require.NoError(t, err)
+
+	value, err := cp.healthCheck(context.Background())
+	require.NoError(t, err)
+	require.Equal(t, int64(2), value)
+}
+
```
asutula, line 59: This is an almost full stack test of the health check, but I'm mocking the `tableland.Tableland` (mesa service) responses.

cmd/healthbot/main.go
```diff
@@ -35,11 +38,24 @@ func main() {
 		if err != nil {
 			log.Fatal().Err(err).Msgf("receipt timeout has invalid format: %s", chainCfg.Probe.ReceiptTimeout)
 		}
+		wallet, err := wallet.NewWallet(chainCfg.Signer.PrivateKey)
+		if err != nil {
+			log.Fatal().Err(err).Msg("unable to create wallet from private key string")
+		}
+
+		chain, err := getChain(chainCfg.Name)
```
asutula, line 46: Getting the chain information to initialize the Go client with from the chain name provided in the config.

cmd/healthbot/main.go
```diff
@@ -56,3 +72,28 @@ func main() {
...
+	case "arbitrum-goerli":
+		return client.Chains.ArbitrumGoerli, nil
+	case "polygon-mumbai":
+		return client.Chains.PolygonMumbai, nil
+	case "local":
+		return client.Chains.Local, nil
+	default:
+		return client.Chain{}, fmt.Errorf("%s is not a valid chain", chain)
+	}
+}
```
asutula, line 99: If we go this route, the chain names used in the config need to be consistent with the names here. These names are what is used in the JS SDK/CLI as well as the Rigs CLI.

internal/router/router.go
```diff
@@ -7,60 +7,29 @@ import (
...
 func ConfiguredRouter(
-	extURLPrefix string,
-	metadataRendererURI string,
+	tableland tableland.Tableland,
+	systemService system.SystemService,
 	maxRPI uint64,
 	rateLimInterval time.Duration,
-	parser parsing.SQLValidator,
-	userStore *user.UserStore,
-	chainStacks map[tableland.ChainID]chains.ChainStack,
```
asutula, line 31: Main change in this file is to make the `ConfiguredRouter` function accept instances of interfaces instead of configuration data to create those instances internally. This is more consistent with good dependency injection design and makes writing tests that use this function better since you can pass in mocks here.

pkg/client/client.go
```diff
@@ -156,15 +156,18 @@ func NewClient(ctx context.Context, wallet *wallet.Wallet, opts ...NewClientOpti
...
+		tblContract, err = ethereum.NewClient(
+			contractBackend,
+			tableland.ChainID(config.chain.ID),
+			config.chain.ContractAddr,
+			wallet,
+			impl.NewSimpleTracker(wallet, contractBackend),
+		)
+		if err != nil {
+			return nil, fmt.Errorf("creating contract client: %v", err)
+		}
```
asutula, line 170: Integrating the client into the healthbot was a good exercise... It made me realize there are times when you want to use the client without a "provider" (alchemy, infura, etc) to make contract calls. This would be cases, like the healthbot, where you're just making read calls and relayed write calls. So now the client can deal with not having a provider/contract backend supplied.

pkg/client/client.go
```diff
@@ -244,6 +247,11 @@ func WithReceiptTimeout(timeout time.Duration) CreateOption {
 
 // Create creates a new table on the Tableland.
 func (c *Client) Create(ctx context.Context, schema string, opts ...CreateOption) (TableID, string, error) {
+	if c.tblContract == nil {
+		return TableID{}, "", errors.New(
+			"clinet created with no provider, must provide an Infura API key, Alchemy API key, or an ETH backend",
+		)
+	}
```
asutula, line 254: For calls that require a provider, we check and error if necessary. Here and below as well.

pkg/client/client.go
```diff
@@ -362,6 +370,11 @@ func (c *Client) Write(ctx context.Context, query string, opts ...WriteOption) (
 		}
 		return res.Transaction.Hash, nil
 	}
+	if c.tblContract == nil {
+		return "", errors.New(
+			"clinet created with no provider, must provide an Infura API key, Alchemy API key, or an ETH backend",
+		)
+	}
```
asutula, line 377: If you expand the code above, you can see this check only happens in the case where we're NOT relaying the write.

pkg/client/client.go
```diff
@@ -515,5 +528,5 @@ func getContractBackend(ctx context.Context, config config) (bind.ContractBacken
 		}
 		return ethclient.DialContext(ctx, url)
 	}
-	return nil, errors.New("no provider specified, must provide an Infura API key, Alchemy API key, or an ETH backend")
+	return nil, nil
```
asutula, line 531: It is no longer an error to not specify a provider.

pkg/client/client_test.go
```diff
@@ -129,184 +109,87 @@ func TestSetController(t *testing.T) {
...
-	require.NoError(t, err)
-
-	err = ep.Start()
-	require.NoError(t, err)
-	t.Cleanup(func() {
-		ep.Stop()
-		client.Close()
-		server.Close()
-	})
+	stack := testutils.CreateFullStack(t, testutils.Deps{})
```
asutula, line 147: Using the general use full stack test harness. Much nicer here.

pkg/testutils/fullstack.go
```diff
@@ -0,0 +1,213 @@
...
+	Parser        parsing.SQLValidator
+	SystemStore   sqlstore.SystemStore
+	UserStore     sqlstore.UserStore
+	ACL           tableland.ACL
+	Tableland     tableland.Tableland
+	SystemService system.SystemService
+}
+
+// CreateFullStack creates the full stack, returning possibly useful things.
+func CreateFullStack(t *testing.T, deps Deps) FullStack {
```
asutula, line 69: The goal here is to provide a general use full stack test harness that also allows you to provide custom implementations or mocks for any of the important dependencies that are defined as `interface` types. If you don't provide a particular dependency via the `deps Deps` arg, a default is used.

There is a "user beware" situation here that I couldn't find a good way around: If you provide a dependency A that was initialized with another instance B that is also represented by `Deps`, you need to pass that instance B in as part of `deps Deps` to make sure the entire dependency tree is consistent (for example, to avoid a situation where `Tableland` is using a mocked version of `UserStore`, but `SystemService` is using the default `UserStore`).

pkg/testutils/fullstack.go
```diff
@@ -0,0 +1,213 @@
...
+type FullStack struct {
+	Backend           *backends.SimulatedBackend
+	Address           common.Address
+	Contract          *ethereum.Contract
+	TransactOpts      *bind.TransactOpts
+	Sk                *ecdsa.PrivateKey
+	Wallet            *wallet.Wallet
+	TblContractClient *ethereum.Client
+	Client            *client.Client
+}
```
asutula, line 55: I just tried to identify all the potentially important things a test may want access to. Me me know if these should change or names aren't clear.

pkg/testutils/fullstack.go
```diff
@@ -0,0 +1,213 @@
+package testutils
```
asutula, line 1: Wasn't really sure where this file should live. Open to suggestions.

pkg/testutils/fullstack.go
brunocalza: Yeah, we have the problem of instantiating dependencies for testing all over the repo. And each place does on it's one way. I think this is good. I can see and future where we have a library with a flexible nice API used in every test (maybe even in main) to create these subcomponents. This can be the beginning of such a library.

brunocalza:
> One realization/question that came out of this (cc @jsign @brunocalza): We're not running healthbot for mainnet networks (that don't support write relay). Do we want to? Now that we're using the Go client, it should just work as long as the account behind the configured private key has a sufficient balance. We'll also need to add an alchemy or infura api key to the `config.json`. The Go client takes care of recognizing whether or not the current chain supports relayed writes or not, and makes the appropriate calls.

The last time I talked with @jsign about this, the take was that we don't want to be spending real money on that. But maybe there's a strong case for it? Or maybe we want to make it so other validators can run healthbot on mainnets if they want to?

asutula:
Hey, yea I was hesitating a bit to merge this one because it requires some major config file changes for anyone running the validator, which always seemed to be a friction point. I was planning on keeping this branch up to date until a good time when we release other changes requiring config file change, but yes, with the new API work, it's probably just a re-write that is needed. Let me know how I can help out.

asutula:
Go ahead and close this if it makes sense. We can look at the closed PR as a reference for some of the dependency and testing stuff.

PR #281, opened 2022-09-23
Adds a quick way of provisioning a validator in GCP

brunocalza:


infra/main.tf
```diff
@@ -0,0 +1,106 @@
...
+variable "credentials_file" {
+    type = string
+}
+
+provider "google" {
+    credentials = file("${var.credentials_file}")
+
+    project = "textile-310716"
+    region  = "us-west1"
+    zone    = "us-west1-b"
```
jsign, line 27: Maybe makes sense to make these also `${var.***}`?

infra/main.tf
```diff
@@ -0,0 +1,106 @@
...
+}
+
+resource "google_compute_instance" "validator" {
+    name         = "${var.vm_name}"
+    machine_type = "e2-standard-2"
+    tags         = ["https-server", "grafana", "ssh-vm"]
+    deletion_protection = false
```
jsign, line 34: In our infra docs we suggest setting this to `true`. 
No strong opinion in doing the same here, just checking if was intentional.

infra/main.tf
```diff
@@ -0,0 +1,106 @@
...
+            image = "ubuntu-2204-jammy-v20220902"
+            type = "pd-ssd"
+            size = "30"
+        }
+    }
+    
+    network_interface {
+        network = "default"
+        access_config {}
+    }
```
jsign, line 47: Prob out of scope for Terraform, but we should remember (maybe comment here?) that the operator should set up firewall rules, but perhaps that's somewhat obvious.

jsign (approved):
Great! :100:

infra/main.tf
brunocalza: We can set the firewall rules here or add a `depends on` so Terraform checks if firewall rule exists

jsign (dismissed):

jsign (approved):

PR #289, opened 2022-09-26
Publishing of local system metrics to a metrics hub server

brunocalza:
# Summary

This PR implements a series of components that allows validators to publish local metrics stored in `metrics.db` to a central metrics hub sever. The server stores metrics in BigQuery, which allows us to have a global view of what is happening in the network. 

# Components included
- A publishing library that fetches unpublished metrics from local db (`metrics.db`) at a regular interval and sends them to an HTTP server
- The implementation of the metrics hub server at `cmd/metricshub`
- The creation of a node id that is stored at a `system_id` table

# Architecture
metrics


docker/deployed/staging/api/config.json
```diff
@@ -41,6 +41,10 @@
       "KeepFiles": 5
     }
   },
+  "TelemetryPublisher": {
```
jsign, line 44: I think we always include all the fields in the `.json` files, so maybe is worth adding `PublishinInterval` even with the default value? (If convinced, this also applies for testnet config json)

pkg/sqlstore/impl/system/internal/db/db.go
```diff
@@ -41,6 +41,12 @@ func Prepare(ctx context.Context, db DBTX) (*Queries, error) {
 	if q.getTablesByStructureStmt, err = db.PrepareContext(ctx, getTablesByStructure); err != nil {
 		return nil, fmt.Errorf("error preparing query GetTablesByStructure: %w", err)
 	}
+	if q.getIdStmt, err = db.PrepareContext(ctx, getId); err != nil {
```
jsign, line 44: FYI, I've in my backlog to create a GH issue to move again to sqlc now that it supports SQLite. Doing this is annoying.

pkg/sqlstore/impl/system/internal/db/id.sql.go
```diff
@@ -0,0 +1,24 @@
+package db
+
+import (
+	"context"
+)
+
+const (
+	getId    = `SELECT id FROM system_id`
+	insertId = `INSERT INTO system_id VALUES (?1)`
```
jsign, line 9: Maybe in the future we'll discover that it's worth having a more general table like `validator_config` and use it as a key-value store. I suspect that will avoid having to create a new table for some new info we need to store.

We can do a migration if that's the case, so fine with this.

cmd/api/main.go
```diff
@@ -208,6 +209,24 @@ func main() {
 		go backupScheduler.Run()
 	}
 
+	var metricsPublisher *publisher.Publisher
+	if config.TelemetryPublisher.Enabled {
+		nodeID, err := chainStacks[config.Chains[0].ChainID].Store.GetID(context.Background())
```
jsign, line 214: Nit: WDYT of doing the `GetID()` call before the L213 `if`, and always print it as `log.Info()`?
I think that might be useful in general to identify logs?

pkg/sqlstore/impl/system_store_instrumented.go
```diff
@@ -393,3 +393,21 @@ func (s *InstrumentedSystemStore) InsertBlockExtraInfo(ctx context.Context, bloc
...
+// GetID returns node identifier.
+func (s *InstrumentedSystemStore) GetID(ctx context.Context) (string, error) {
+	start := time.Now()
+	id, err := s.store.GetID(ctx)
+	latency := time.Since(start).Milliseconds()
+
+	attributes := []attribute.KeyValue{
+		{Key: "method", Value: attribute.StringValue("Id")},
+		{Key: "success", Value: attribute.BoolValue(err == nil)},
+		{Key: "chainID", Value: attribute.Int64Value(int64(s.chainID))},
```
jsign, line 406: This could be avoided since might be noisy for filtering logs per chain considering it's irrelevant.
Some side effects of not having the split store. (nre)

pkg/telemetry/publisher/publisher.go
```diff
@@ -0,0 +1,107 @@
...
+		return nil
+	}
+
+	if err := p.exporter.Export(ctx, metrics, p.nodeID); err != nil {
+		return fmt.Errorf("export metrics: %s", err)
+	}
+
+	rowsIds := make([]int64, len(metrics))
+	for i, m := range metrics {
+		rowsIds[i] = m.RowID
```
jsign, line 88: `rowid` to the rescue.

pkg/telemetry/publisher/publisher.go
```diff
@@ -0,0 +1,107 @@
...
+
+// NewPublisher creates a new publisher.
+func NewPublisher(s MetricsStore, e MetricsExporter, nodeID string, interval time.Duration) *Publisher {
+	return &Publisher{
+		store:    s,
+		exporter: e,
+
+		nodeID:      nodeID,
+		interval:    interval,
+		fetchAmount: 100,
```
jsign, line 34: Good idea. Fine with having this fixed.

pkg/telemetry/storage/db.go
```diff
@@ -73,6 +77,85 @@ func (db *TelemetryDatabase) StoreMetric(ctx context.Context, metric telemetry.M
...
+	}
+
+	rowsAffected, err := result.RowsAffected()
+	if err != nil {
+		return fmt.Errorf("rows affected: %s", err)
+	}
+
+	if rowsAffected != int64(len(rowids)) {
+		return fmt.Errorf("rows affected %d, differs from rowids length %d", rowsAffected, int64(len(rowids)))
+	}
```
jsign, line 154: :+1:

jsign (approved):
LGTM, nice!

cmd/api/main.go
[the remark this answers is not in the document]
brunocalza: Agree. Splitting would be really good 👍

brunocalza: Sure!

PR #322, opened 2022-10-03
otel: update breaking changing apis & other deps

jsign:
# Summary

This PR updates the OTEL dependencies, which have breaking APIs, and other deps.

# Context

We were notified by dependabot that new dependency versions were available.

# Implementation overview

The OTEL breaking changes were a bit annoying since there's no documentation on how to fix them. 
I had to do some GH search and code reading to figure out how to fix this properly. 

I slightly modified the bucket numbers to have extra buckets between the 10ms and 100ms range.


jsign:
I haven't closed this PR explicitly. Looks like GH has some rules if you revert changes by doing some automatic closing? :shrug: 
In any case, I'll re-open soon.

PR #323, opened 2022-10-03
otel: update breaking changing apis

jsign:
# Summary

This PR updates the OTEL dependencies, which have breaking APIs, and other deps.

**Note**: This PR will be a draft until https://github.com/open-telemetry/opentelemetry-go/issues/3229 gets fixed in the next OTEL version. This is blocking because `otelsql` relies on OTEL deduping re-registering metrics; if that invariant doesn't hold, it causes a panic. We'll wait until there's a new OTEL version, and see how everything works.

# Context

We were notified by dependabot that new dependency versions were available.

# Implementation overview

The OTEL breaking changes were a bit annoying since there's no documentation on how to fix them. 
I had to do some GH search and code reading to figure out how to fix this properly. 

I slightly modified the bucket numbers to have extra buckets between the 10ms and 100ms range.


brunocalza (approved):
💯

PR #325, opened 2022-10-04
Adds the possibility to start a validator from a backup file

### LINKED ISSUE #303: Allow a fresh validator to sync from a snapshot URL
Our cloud validator is creating snapshots of its database with some frequency.
This is useful for many things, but in particular, it can be the basis to allow a validator to sync from a trusted/verified point in time forward.

Some obvious use-cases for this:
- Spinning a second cloud validator quickly.
- Allowing external validators to have a synced validator within less time and with fewer costs than syncing from scratch.
- Have a local validator up and synced as fast as possible for local debugging.

I'm pretty sure we need to do some extra work in the backup databases since they contain data in some tables that shouldn't be there. For example, we still copy the relayed transactions table, etc.

As part of uploading the database backups to some cloud storage to have a public URL, we can have a previous step of cleaning up information that shouldn't be there. But we can also fine-tune the backup process to generate this database directly.

Open to comments, it might require some more dedicated scoping.

brunocalza:
# Summary

Adds the option to start a validator with a database restored from a backup file.

# Context

As the number of transactions in Tableland grows the need for a fast spin-up of a fully synced validator becomes more evident. Also, as we found more cases of non-determinism we have to ask node operators to resync their state, and having a fast way of doing that from a backup file speed things up and reduces costs. 

Closes #303 

# Implementation overview

An option to the Makefile was added so you can choose to start a validator with a database restored from a backup file URL. 

The user can provide a backup URL to indicate it wants to bootstrap the validator from a backup database by running `BOOTSTRAP_BACKUP_URL=https://storage.googleapis.com/tableland-testnet/backups/tbl_backup_latest.db.zst make testnet-up`

That will download the latest backup file from Cloud Storage, decompress it, and move the file to the right location. The validator will start from the point that backup the database left off. 

- A new package in `backup` package called `restorer` was created. The component called `BackupRestored` is responsible for the process of downloading the file and loading the database. 
- The backup file is stored in Google Cloud storage in an open (only for reads) bucket. Each environment has its own bucket. 
- The latest backup file is named `tbl_backup_latest.db.zst` 
- We haven't implemented the upload process yet. But the idea is that every time we have a new backup file we upload it with its original name (that has the timestamp) and make a copy with the `tbl_backup_latest.db.zst` name. So `tbl_backup_latest.db.zst` will always point to the latest. 
- The restorer does not make any modifications to the backup database (except from deleting `system_pending_tx` rows)
- The `main.go` checks `BOOTSTRAP_BACKUP_URL` config and passes it to the restorer.

# Checklist


cmd/api/config.go
```diff
@@ -54,13 +54,15 @@ type config struct {
 			Enabled   bool `default:"true"`
 			KeepFiles int  `default:"5"` // number of files to keep
 		}
+		URL string `default:""`
```
brunocalza, line 57: This is the backup file url. Not sure if it makes sense to have a default value here.

docker/Makefile
```diff
@@ -16,6 +16,7 @@ staging-up:
 	USER=$(shell id -u) \
 	ENVIRONMENT=deployed/staging \
 	COMPOSE_DOCKER_CLI_BUILD=1 \
+	FROM_BACKUP=$(shell if [ ${FROM_BACKUP} = "true" ]; then echo "true"; else echo "false"; fi) \
```
brunocalza, line 19: Adjustments to Makefile to enable `FROM_BACKUP` option

docker/deployed/staging/api/config.json
```diff
@@ -39,7 +39,8 @@
     "Pruning": {
       "Enabled": true,
       "KeepFiles": 5
-    }
+    },
+    "URL" : "https://storage.googleapis.com/tableland-staging/backups/tbl_backup_latest.db.zst"
```
brunocalza, line 43: Points to the latest staging backup file. I uploaded this manually. Later will have an automatic process.

docker/deployed/testnet/api/config.json
```diff
@@ -40,7 +40,8 @@
         "Pruning": {
             "Enabled": true,
             "KeepFiles": 5
-        }
+        },
+        "URL" : "https://storage.googleapis.com/tableland-testnet/backups/tbl_backup_latest.db.zst"
```
brunocalza, line 44: Points to the latest testnet backup file.

pkg/backup/compressor.go
```diff
@@ -59,3 +60,46 @@ func Compress(filepath string) (string, error) {
 	}
 	return newFilepath, nil
 }
+
+// Decompress decompresses a zstd file.
+func Decompress(filepath string) (string, error) {
```
brunocalza, line 65: Implementation of decompression

pkg/backup/restorer/restorer.go
```diff
@@ -0,0 +1,110 @@
...
+// NewBackupRestorer creates a new BackupRestorer.
+func NewBackupRestorer(url, dst string) *BackupRestorer {
+	return &BackupRestorer{
+		url: url,
+		dst: dst,
+	}
+}
+
+// Restore restores a database from a backup file URL.
+func (br *BackupRestorer) Restore() error {
```
brunocalza, line 26: This is a very straightforward approach: download the file, decompress, and rename it. Lmk if this is good enough for now or we need to add extras

jsign (approved):
LGTM! :muscle: 
I left some comments around configs to get your thoughts.

cmd/api/main.go
```diff
@@ -57,6 +58,22 @@ func main() {
...
+			log.Fatal().
+				Err(err).
+				Msg("backup restoration failed")
+		}
+		log.Info().Float64("elapsed_time", time.Since(elapsedTime).Seconds()).Msg("backup restore finished")
+	}
+
 	databaseURL := fmt.Sprintf(
 		"file://%s?_busy_timeout=5000&_foreign_keys=on&_journal_mode=WAL",
 		path.Join(dirPath, "database.db"),
```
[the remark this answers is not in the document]
brunocalza: yeah. much better. thanks 👍

pkg/backup/restorer/restorer.go
```diff
@@ -0,0 +1,110 @@
...
+	}
+
+	_, err := backup.Decompress(fmt.Sprintf("%s/backup.db.zst", br.dst))
+	if err != nil {
+		return fmt.Errorf("decompress: %s", err)
+	}
+
+	if err := br.load(); err != nil {
+		return fmt.Errorf("loading the database: %s", err)
+	}
```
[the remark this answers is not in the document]
brunocalza: good idea!

cmd/api/config.go
```diff
@@ -54,13 +54,15 @@ type config struct {
...
 	}
 	TelemetryPublisher struct {
 		Enabled            bool   `default:"false"`
 		MetricsHubURL      string `default:""`
 		MetricsHubAPIKey   string `default:""`
 		PublishingInterval string `default:"10s"`
 	}
+	FromBackup bool `default:"false" env:"FROM_BACKUP"`
```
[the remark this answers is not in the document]
jsign: > I've started with only one config, much like `BoostrapFromSnapshotURL`, but the problem with that is:
> 
> we can't commit a configured value in `config.json` for that config, because that would mean that in every deployment the validator would be bootstrapped from backup. Or the user would have to change the `config.json`, which is bad because `config.json` is committed.
> 
> This implies that the user would have to provide the URL every time he wanted to bootstrap a validator from a backup file. I thought that it is just bad UX. So I aimed for something simpler like: `FROM_BACKUP=true make testnet-up`.
>
> Does that make sense to you?

Yes, that's what I suspected with my comment:
> I think you landed on the current configuration since you wanted a way to have a true and false UX with an "opinionated" preconfigured URL at the config.json.

I'm not sure about asking them to set the URL to be bad UX. My view on this is that it is more like "acceptable" or even "transparent" UX.

We should remember that we're aiming to build a decentralized network, where validators shouldn't be strictly/blindly trusting anyone about the state of the network. When a validator syncs using a backup, they're indirectly trusting the source that it has built the state correctly. (Unless we have a way to validate snapshots if Tableland state roots are posted on-chain, which we might have in the future but not today)

To be clear, I think it's fine to leave the pre-configured URL. I'm just dropping some extra thoughts about how a default URL can bake trust assumptions. In any case, validators should know what it means to boostrap from a backup even if the validator has preconfigured values. My rambling here is mostly philosophical and semantics. :)

> I agree that grouping them is much better. This would lead to something like:

:+1: 

> (side note: I'm insisting on the word backup, instead of snapshot, just to be consistent with the code and ultimately with SQLIte Backup API)

Coo, no strong opinion here. I usually use "snapshot" since it's usually the word used in many other validators out there. 
I think "backup" sounds somewhat "personal", and "snapshot" is more agnostic. Since we have a "backup" we need this extra logic when loading it to "remove data" to make it agnostic. Again, maybe a delicatessen here but I'm fine with backup.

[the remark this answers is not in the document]
brunocalza: Makes sense. I haven't looked for that angle. Also, it would probably make it easier for us debugging since we could choose a backup from a specific timestamp. 

Changing to:
`BOOTSTRAP_BACKUP_URL=https://storage.googleapis.com/tableland-staging/backups/tbl_backup_latest.db.zst make staging-up`

PR #328, opened 2022-10-05
Fix WAL checkpoint starvation due to unclosed migration connections

### LINKED ISSUE #312: Investigate WAL checkpointing problem
We've detected that the WAL size of the validator seems to be growing without bounds. 

This might be caused by Checkpointing starvation as described in the [SQLite WAL documentation](https://sqlite.org/wal.html). After a quick look some time ago, we haven't found any obvious reason why this would be the case. We had even set all the database pools to `SetMaxIdleConnections(0)` to be sure idle connections were interpreted as "open" thus blocking checkpoints.

Unfortunately, being so aggressive with idle connections has a considerable performance hit in all the validators. The performance hit isn't noticeable today due to our needed throughput. But, considering it hasn't solved the checkpointing problem, it should go away.

We need to dedicate some extra time to check what might be happening regarding checkpoint starvation to see if we see the ultimate cause. If we can't find it, we should create a background process in the validator that [forces checkpointing via a PRAGMA](https://www.sqlite.org/c3ref/wal_checkpoint_v2.html).

jsign:
# Summary

This PR fixes a bug where we didn't properly close database connections used for database migrations.
I've checked in my local validator and tested that with/without this change, my WAL file does/doesn't get checkpointed.

I've also removed the `MaxIdleConns(0)`, which always forces us to open a new connection, which we've seen has a considerable hit on the overall performance of the validator. Let's get back to square one now that this _should_ be fixed and stop paying a performance penalty that didn't even fixed/alleviated the problem.

Note that we're trying to avoid as much as possible having any background goroutine to force checkpointing and let the natural behavior of SQLite be enough. Considering we found the source of the problem, we'll see now if that's the case.

Closes #312.

# Context

See #312.

# Implementation overview

The current fix and other changes don't require much explanation. The way I found this was by spinning up partially wired validators, and checking locally if the WAL gets checkpointed. The interesting part was that the graceful closing of the main wiring was correct, so _apparently_ all looked fine. After a while, I landed on the side-channel db pool created for migrations.

There was an interesting side-effect of fixing this bug on tests.

After fixing this bug, I saw most tests fail since expected system tables _didn't exist_. This might sound surprising at first, but it makes sense (and is even a further validation that this fix works). In memory, databases in SQLite get deleted when the last open connection closes. When we spin up test stacks, we don't have "warmed up" any db connection pool, so before running the tests, there aren't open connections; thus, the in-memory `dbURI` is destroyed.

When we had the bug, this didn't happen since our system store (or other deps) ran table migrations. And these migrations weren't closing their database connection, so at least there was always one active database connection preventing the in-memory database from being destroyed.

The fix for this is quite simple and means always forcing a single connection to be opened in every in-memory database so it doesn't get destroyed if there are no active database connections at any point in time. See the PR comments for more info.

Not required.

# Checklist

### DIFF
--- a/tests/sqlite3.go
+++ b/tests/sqlite3.go
@@ -1,10 +1,25 @@
 package tests
 
 import (
+	"context"
+	"database/sql"
+	"testing"
+
 	"github.com/google/uuid"
+	"github.com/stretchr/testify/require"
 )
 
 // Sqlite3URI returns a URI to spinup an in-memory Sqlite database.
-func Sqlite3URI() string {
-	return "file::" + uuid.NewString() + ":?mode=memory&cache=shared&_foreign_keys=on"
+func Sqlite3URI(t *testing.T) string {
+	dbURI := "file::" + uuid.NewString() + ":?mode=memory&cache=shared&_foreign_keys=on"
+	db, err := sql.Open("sqlite3", dbURI)
+	require.NoError(t, err)
+	conn, err := db.Conn(context.Background())
+	require.NoError(t, err)
+	t.Cleanup(func() {
+		conn.Close()
+		db.Close()
+	})
+
+	return dbURI
 }
--- a/tests/sqlite3.go
+++ b/tests/sqlite3.go
@@ -17,8 +17,8 @@ func Sqlite3URI(t *testing.T) string {
 	conn, err := db.Conn(context.Background())
 	require.NoError(t, err)
 	t.Cleanup(func() {
-		conn.Close()
-		db.Close()
+		_ = conn.Close()
+		_ = db.Close()
 	})
 
 	return dbURI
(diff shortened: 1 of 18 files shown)


tests/sqlite3.go
```diff
@@ -1,10 +1,25 @@
...
+	db, err := sql.Open("sqlite3", dbURI)
+	require.NoError(t, err)
+	conn, err := db.Conn(context.Background())
+	require.NoError(t, err)
+	t.Cleanup(func() {
+		_ = conn.Close()
+		_ = db.Close()
+	})
+
+	return dbURI
```
jsign, line 24: This change is needed to avoid in-memory SQLite databases being unintentionally deleted since we don't have active connections.

What we do is create a database pool and ask for a connection. When then, register a function in `t.Cleanup(...)` to close the connection and the pool when the tests finish.

brunocalza (approved):
Excellent! 💯

PR #332, opened 2022-10-06
Adds animation_url to table's metadata

### LINKED ISSUE #326: Add animation_url to table NFT uri object
I'm working on an HTML app which will render a Table's contents, along with the query (which will be editable by the user). 
The HTML app will pretty simple. There will be an input box for a query, which can be edited, but will be prepopulated with a `SELECT * FROM ${table_from_page_params} LIMIT 20`, and that query will be automatically run, with the contents shown below the query box. The animation_url of a Table's NFT would point to this app. Eventually, I'm thinking it could be an IPFS pointer, but for now we can point to it the same way we point to the SVG at render.tableland.xyz. 

For now, I'm thinking `https://render.tableland.xyz/anim/?chain=1&id=1`,

brunocalza:
# Summary

Adds `animation_url` property to the response of Table's metadata call `GET /chain/{chainID}/tables/{id}`

Closes #326 

cc @awmuncy 

# Context

An HTML app (light version of the dashboard) pointed by `animation_url` address will be rendered on most NFT marketplaces when the Table NFT details page is opened. 

# Implementation overview

Adds a new config `AnimationRendererURI`. In `testnet` it points to `https://render.tableland.xyz/anim/`. In other envs it's just an empty string, which means the `animation_url` won't be shown. 

n/a

# Checklist

### DIFF
--- a/cmd/api/config.go
+++ b/cmd/api/config.go
@@ -21,9 +21,8 @@ type config struct {
 	Dir                string // This will default to "", NOT the default dir value set via the flag package
 	BootstrapBackupURL string `default:"" env:"BOOTSTRAP_BACKUP_URL"`
 
-	HTTP    HTTPConfig
-	Gateway GatewayConfig
-
+	HTTP             HTTPConfig
+	Gateway          GatewayConfig
 	TableConstraints TableConstraints
 	QueryConstraints QueryConstraints
 
@@ -56,8 +55,9 @@ type HTTPConfig struct {
 
 // GatewayConfig contains configuration for the Gateway.
 type GatewayConfig struct {
-	ExternalURIPrefix   string `default:"https://testnet.tableland.network"`
-	MetadataRendererURI string `default:""`
+	ExternalURIPrefix    string `default:"https://testnet.tableland.network"`
+	MetadataRendererURI  string `default:""`
+	AnimationRendererURI string `default:""`
 }
 
 // BackupConfig contains configuration for automatic database backups.
--- a/docker/deployed/staging/api/config.json
+++ b/docker/deployed/staging/api/config.json
@@ -8,7 +8,8 @@
   },
   "Gateway": {
     "ExternalURIPrefix": "https://staging.tableland.network",
-    "MetadataRendererURI": ""
+    "MetadataRendererURI": "",
+    "AnimationRendererURI": ""
   },
   "DB": {
     "Port": "5432"
--- a/docker/deployed/testnet/api/config.json
+++ b/docker/deployed/testnet/api/config.json
@@ -9,7 +9,8 @@
     },
     "Gateway": {
         "ExternalURIPrefix": "https://testnet.tableland.network",
-        "MetadataRendererURI": "https://render.tableland.xyz"
+        "MetadataRendererURI": "https://render.tableland.xyz",
+        "AnimationRendererURI": "https://render.tableland.xyz/anim"
     },
     "DB": {
         "Port": "5432"
--- a/internal/system/impl/sqlstore.go
+++ b/internal/system/impl/sqlstore.go
@@ -30,13 +30,17 @@ const (
 
 	// DefaultMetadataImage is the default image for table's metadata.
 	DefaultMetadataImage = "https://bafkreifhuhrjhzbj4onqgbrmhpysk2mop2jimvdvfut6taiyzt2yqzt43a.ipfs.dweb.link"
+
+	// DefaultAnimationURL is an empty string. It means that the attribute will not appear in the JSON metadata.
+	DefaultAnimationURL = ""
 )
 
 // SystemSQLStoreService implements the SystemService interface using SQLStore.
 type SystemSQLStoreService struct {
-	extURLPrefix        string
-	metadataRendererURI string
-	stores              map[tableland.ChainID]sqlstore.SystemStore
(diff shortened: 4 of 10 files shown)


cmd/api/config.go
```diff
@@ -28,8 +28,9 @@ type config struct {
 		MaxRequestPerInterval uint64 `default:"10"`
 	}
 	Gateway struct {
-		ExternalURIPrefix   string `default:"https://testnet.tableland.network"`
-		MetadataRendererURI string `default:""`
+		ExternalURIPrefix    string `default:"https://testnet.tableland.network"`
+		MetadataRendererURI  string `default:""`
+		AnimationRendererURI string `default:""`
```
brunocalza, line 33: Although `AnimationRendererURI ` and `MetadataRendererURI` point to the same server, I added a new config because in the future they will not.

docker/deployed/staging/api/config.json
```diff
@@ -8,7 +8,8 @@
   },
   "Gateway": {
     "ExternalURIPrefix": "https://staging.tableland.network",
-    "MetadataRendererURI": ""
+    "MetadataRendererURI": "",
+    "AnimationRendererURI": ""
```
brunocalza, line 12: We don't have an app to render `staging` data, this means the value is set to an empty string. This implies the response of the `GET /chain/{chainID}/tables/{id}` call won't include the `animation_url` property. See [omitempty](https://github.com/tablelandnetwork/go-tableland/pull/332/files#diff-6ca5d243b7947601125912e4889dbdd233c1c22aefcedee8b18a10e83e972e1bR45) in the the struct definition.

docker/deployed/testnet/api/config.json
```diff
@@ -9,7 +9,8 @@
     },
     "Gateway": {
         "ExternalURIPrefix": "https://testnet.tableland.network",
-        "MetadataRendererURI": "https://render.tableland.xyz"
+        "MetadataRendererURI": "https://render.tableland.xyz",
+        "AnimationRendererURI": "https://render.tableland.xyz/anim"
```
brunocalza, line 13: Testnet URL as mentioned  in #326

internal/system/impl/sqlstore.go
```diff
@@ -180,6 +187,20 @@ func (s *SystemSQLStoreService) getMetadataImage(chainID tableland.ChainID, tabl
...
+		return DefaultAnimationURL
+	}
+
+	uri := strings.TrimRight(s.animationRendererURI, "/")
+	if _, err := url.ParseRequestURI(uri); err != nil {
+		log.Warn().Str("uri", uri).Msg("metadata renderer uri could not be parsed")
+		return DefaultAnimationURL
+	}
+
+	return fmt.Sprintf("%s/?chain=%d&id=%s", uri, chainID, tableID)
```
brunocalza, line 201: The pattern as mentioned in #326

internal/system/impl/sqlstore.go
```diff
@@ -180,6 +187,20 @@ func (s *SystemSQLStoreService) getMetadataImage(chainID tableland.ChainID, tabl
...
+func (s *SystemSQLStoreService) getAnimationURL(chainID tableland.ChainID, tableID tables.TableID) string {
+	if s.animationRendererURI == "" {
+		return DefaultAnimationURL
+	}
+
+	uri := strings.TrimRight(s.animationRendererURI, "/")
+	if _, err := url.ParseRequestURI(uri); err != nil {
+		log.Warn().Str("uri", uri).Msg("metadata renderer uri could not be parsed")
+		return DefaultAnimationURL
+	}
```
jsign, line 199: Maybe we can do this check in the ctor, and fail fast signaling that the configuration of the validator is incorrect instead of defaulting to a global var (which probably wasn't the intention if a configuration was provided).

Also, `uri` would have been precalculated so we don't need to do this parsing on every call.

jsign (approved):
LGTM! I've mentioned a small nit for your consideration.

awmuncy (dismissed):
Looks good!

internal/system/impl/sqlstore.go
brunocalza: yeah. moving over there

internal/system/impl/sqlstore.go
brunocalza: `tableID` has the `big.Int` type. so it must be converted to `string`

internal/system/impl/sqlstore.go
```diff
@@ -30,28 +30,49 @@ const (
...
 func NewSystemSQLStoreService(
 	stores map[tableland.ChainID]sqlstore.SystemStore,
 	extURLPrefix string,
 	metadataRendererURI string,
+	animationRendererURI string,
 ) (system.SystemService, error) {
 	if _, err := url.ParseRequestURI(extURLPrefix); err != nil {
 		return nil, fmt.Errorf("invalid external url prefix: %s", err)
 	}
+
```
brunocalza, line 56: @jsign moved both uris validations to here

jsign: SGTM

awmuncy:
Hey @brunocalza, the work on the renderer is ready to be released. So we can merge this when you are ready.

jsign (approved):

Issue #302, opened 2022-09-30
Consider re-organizing dependencies creation


jsign:
I think our `main` function is reasonably clear, but we can probably do the best job organizing how the dependency tree gets created. 
We don't need to use any dependency injection tool or magical indirections. 

I'm referring to better encapsulation, making it more obvious that everything that should be gracefully close is registered, and avoiding cognitive load in figuring out how the modules fit together.

There's no rush to do this, I'm capturing a feeling.

PR #334, opened 2022-10-07
Telemetry: add binary git summary & last executed block number per chain id

jsign:
# Summary

This PR adds two new metrics to the optional telemetry collection:
- The Git summary of the validator binary: This will help understand if validators are running old versions.
- Last executed block number per chain: this will help to understand if validators have some healthiness problem syncing chains, lagging behind; we could infer syncing speeds or liveness since the metric is published at a fixed time-frequency.

This was tested in staging, and it's working correctly.

# Context

This is part of the objective to add optional telemetry reporting in validators.

# Implementation overview

The implementation has two main parts:
1) Git summary metrics: this is published once when the validator spins up. 
2) Last executed block number: this is done by a new telemetry sub-package, collecting the information in all `chainStacks`, and aggregating them in a single metric to be published.

No specific directions are needed. I'll add some PR comments to mention some tangential refactors.

# Checklist

### DIFF
--- a/buildinfo/buildinfo.go
+++ b/buildinfo/buildinfo.go
@@ -38,9 +38,20 @@ func GetSummary() Summary {
 	return summary
 }
 
-func (s Summary) GetGitCommit() string     { return s.GitCommit }
-func (s Summary) GetGitBranch() string     { return s.GitBranch }
-func (s Summary) GetGitState() string      { return s.GitState }
-func (s Summary) GetGitSummary() string    { return s.GitSummary }
-func (s Summary) GetBuildDate() string     { return s.BuildDate }
+// GetGitCommit returns the GitCommit.
+func (s Summary) GetGitCommit() string { return s.GitCommit }
+
+// GetGitBranch returns the GitBranch.
+func (s Summary) GetGitBranch() string { return s.GitBranch }
+
+// GetGitState returns the GitState.
+func (s Summary) GetGitState() string { return s.GitState }
+
+// GetGitSummary returns the GitSummary.
+func (s Summary) GetGitSummary() string { return s.GitSummary }
+
+// GetBuildDate returns the build date.
+func (s Summary) GetBuildDate() string { return s.BuildDate }
+
+// GetBinaryVersion returns the binary version.
 func (s Summary) GetBinaryVersion() string { return s.Version }
--- a/buildinfo/buildinfo.go
+++ b/buildinfo/buildinfo.go
@@ -25,7 +25,7 @@ type Summary struct {
 	Version    string
 }
 
-// GetSummary returns a JSON string with a summary of git information.
+// GetSummary returns a summary of git information.
 func GetSummary() Summary {
 	summary := Summary{
 		GitCommit:  GitCommit,
@@ -37,3 +37,10 @@ func GetSummary() Summary {
 	}
 	return summary
 }
+
+func (s Summary) GetGitCommit() string     { return s.GitCommit }
+func (s Summary) GetGitBranch() string     { return s.GitBranch }
+func (s Summary) GetGitState() string      { return s.GitState }
+func (s Summary) GetGitSummary() string    { return s.GitSummary }
+func (s Summary) GetBuildDate() string     { return s.BuildDate }
+func (s Summary) GetBinaryVersion() string { return s.Version }
--- a/cmd/api/config.go
+++ b/cmd/api/config.go
@@ -55,12 +55,14 @@ type config struct {
 			KeepFiles int  `default:"5"` // number of files to keep
 		}
 	}
-	TelemetryPublisher struct {
-		Enabled            bool   `default:"false"`
-		MetricsHubURL      string `default:""`
-		MetricsHubAPIKey   string `default:""`
-		PublishingInterval string `default:"10s"`
-	}
+	TelemetryPublisher TelemetryPublisherConfig
+}
+
+type TelemetryPublisherConfig struct {
+	Enabled            bool   `default:"false"`
+	MetricsHubURL      string `default:""`
+	MetricsHubAPIKey   string `default:""`
+	PublishingInterval string `default:"10s"`
 }
 
 // TableConstraints describes contraints to be enforced for Tableland tables.
--- a/cmd/api/config.go
+++ b/cmd/api/config.go
@@ -63,6 +63,8 @@ type TelemetryPublisherConfig struct {
 	MetricsHubURL      string `default:""`
 	MetricsHubAPIKey   string `default:""`
 	PublishingInterval string `default:"10s"`
+
+	ChainStackCollectFrequency string `default:"15m"`
 }
 
 // TableConstraints describes contraints to be enforced for Tableland tables.
--- a/cmd/api/config.go
+++ b/cmd/api/config.go
@@ -58,6 +58,7 @@ type config struct {
 	TelemetryPublisher TelemetryPublisherConfig
 }
 
+// TelemetryPublisherConfig contains configuration attributes for the telemetry module.
 type TelemetryPublisherConfig struct {
 	Enabled            bool   `default:"false"`
 	MetricsHubURL      string `default:""`
--- a/cmd/api/main.go
+++ b/cmd/api/main.go
@@ -456,7 +456,7 @@ func configureTelemetry(
 	}()
 
 	// Module closing function
-	close := func(ctx context.Context) error {
+	return func(ctx context.Context) error {
 		clsChainStackCollector()
 		<-chainStackCollectorClosed
 
@@ -468,7 +468,5 @@ func configureTelemetry(
 		}
 
 		return nil
-	}
-
-	return close, nil
+	}, nil
 }
--- a/cmd/api/main.go
+++ b/cmd/api/main.go
@@ -45,6 +45,8 @@ import (
 	"go.opentelemetry.io/otel/attribute"
 )
 
+type moduleClose func(ctx context.Context) error
+
 func main() {
 	config, dirPath := setupConfig()
 	logging.SetupLogger(buildinfo.GitCommit, config.Log.Debug, config.Log.Human)
@@ -114,18 +116,6 @@ func main() {
 	if err != nil {
 		log.Fatal().Err(err).Msg("creating user store")
 	}
-
-	metricsDatabaseURL := fmt.Sprintf(
-		"file://%s?_busy_timeout=5000&_foreign_keys=on&_journal_mode=WAL",
-		path.Join(dirPath, "metrics.db"),
-	)
-
-	metricsStore, err := storage.New(metricsDatabaseURL)
-	if err != nil {
-		log.Fatal().Err(err).Msg("creating metrics store")
-	}
-
-	telemetry.SetMetricStore(metricsStore)
 
 	rateLimInterval, err := time.ParseDuration(config.HTTP.RateLimInterval)
 	if err != nil {
@@ -232,21 +222,9 @@ func main() {
 
 	log.Info().Str("node_id", nodeID).Msg("node info")
 
-	var metricsPublisher *publisher.Publisher
-	if config.TelemetryPublisher.Enabled {
-		exporter, err := publisher.NewHTTPExporter(
-			config.TelemetryPublisher.MetricsHubURL,
-			config.TelemetryPublisher.MetricsHubAPIKey,
-		)
-		if err != nil {
-			log.Fatal().Err(err).Msg("creating metrics http exporter")
-		}
-		publishingInterval, err := time.ParseDuration(config.TelemetryPublisher.PublishingInterval)
-		if err != nil {
-			log.Fatal().Err(err).Msg("parsing publising interval")
-		}
-		metricsPublisher = publisher.NewPublisher(metricsStore, exporter, nodeID, publishingInterval)
-		metricsPublisher.Start()
+	closeTelemetryModule, err := configureTelemetry(nodeID, dirPath, config.TelemetryPublisher)
+	if err != nil {
+		log.Fatal().Err(err).Msg("configuring telemetry")
 	}
 
 	cli.HandleInterrupt(func() {
@@ -279,15 +257,12 @@ func main() {
 			log.Error().Err(err).Msg("closing user store")
 		}
 
-		if err := metricsStore.Close(); err != nil {
-			log.Error().Err(err).Msg("closing metrics store")
-		}
-
-		if config.TelemetryPublisher.Enabled {
-			metricsPublisher.Close()
-		}
 		if err := executorsDB.Close(); err != nil {
 			log.Error().Err(err).Msg("closing executors db")
+		}
+
+		if err := closeTelemetryModule(ctx); err != nil {
+			log.Error().Err(err).Msg("closing telemetry module")
 		}
 	})
 }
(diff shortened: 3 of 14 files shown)


buildinfo/buildinfo.go
```diff
@@ -37,3 +37,21 @@ func GetSummary() Summary {
...
+func (s Summary) GetGitState() string { return s.GitState }
+
+// GetGitSummary returns the GitSummary.
+func (s Summary) GetGitSummary() string { return s.GitSummary }
+
+// GetBuildDate returns the build date.
+func (s Summary) GetBuildDate() string { return s.BuildDate }
+
+// GetBinaryVersion returns the binary version.
+func (s Summary) GetBinaryVersion() string { return s.Version }
```
jsign, line 57: I took advantage of the existing `Summary` struct, and made it implement the new corresponding `telemetry.GitSummary` interface.

buildinfo/buildinfo.go
```diff
@@ -25,7 +25,7 @@ type Summary struct {
 	Version    string
 }
 
-// GetSummary returns a JSON string with a summary of git information.
+// GetSummary returns a summary of git information.
```
jsign, line 28: Nit: fixing a wrong comment.

cmd/api/config.go
```diff
@@ -55,12 +55,17 @@ type config struct {
 			KeepFiles int  `default:"5"` // number of files to keep
 		}
 	}
-	TelemetryPublisher struct {
-		Enabled            bool   `default:"false"`
-		MetricsHubURL      string `default:""`
-		MetricsHubAPIKey   string `default:""`
-		PublishingInterval string `default:"10s"`
-	}
+	TelemetryPublisher TelemetryPublisherConfig
```
jsign, line 58: I extracted the struct to be fully-defined. It will become clear why in `main` comments.

cmd/api/config.go
```diff
@@ -55,12 +55,17 @@ type config struct {
...
+}
+
+// TelemetryPublisherConfig contains configuration attributes for the telemetry module.
+type TelemetryPublisherConfig struct {
+	Enabled            bool   `default:"false"`
+	MetricsHubURL      string `default:""`
+	MetricsHubAPIKey   string `default:""`
+	PublishingInterval string `default:"10s"`
+
+	ChainStackCollectFrequency string `default:"15m"`
```
jsign, line 68: This `ChainStackCollectFrequency` is a new configuration. It specifies how often the last executed block number per chain is collected. The name is more generic since the same telemetry sub-module can be used to also collect more info about each chain-stack.

cmd/api/main.go
```diff
@@ -39,12 +39,15 @@ import (
 	"github.com/textileio/go-tableland/pkg/sqlstore/impl/user"
```
jsign, line 39: A big chunk of what happened in this file is a step forward in the https://github.com/tablelandnetwork/go-tableland/issues/302 direction.

I believe that our `main` function is a bit convoluted, so what I did here is a semi-proposal experiment to show if the same strategy could be done iteratively in other parts.

I encapsulated all the wiring related to `telemetry` in a single function, and only call that in `main`, which:
1) Makes it clear what is needed to wire the "telemetry module". There's no need to understand how other wires are related to telemetry.
2) The function returns a `moduleClose` (L49) that should be used by `main` to gracefully close _everything_ related to telemetry. This helps `main` to have a clear rule of always knowing that it should call returned `moduleClose` functions when the process is closed. 
3) Point 2) also helps the `cli.HandleInterrupt(...)` logic be clearer. Instead of closing fine-grained moving parts, it closes "higher-order" modules (which were built by the corresponding wiring function which knows what is needed, not "main").

The result should be a shorter, high-level, and clearer `main` function. 
I applied this idea only with the `telemetry` module since it's related to this PR. Also, to see if this starts to feel convincing.

cmd/api/main.go
```diff
@@ -232,21 +223,9 @@ func main() {
...
-		publishingInterval, err := time.ParseDuration(config.TelemetryPublisher.PublishingInterval)
-		if err != nil {
-			log.Fatal().Err(err).Msg("parsing publising interval")
-		}
-		metricsPublisher = publisher.NewPublisher(metricsStore, exporter, nodeID, publishingInterval)
-		metricsPublisher.Start()
+	closeTelemetryModule, err := configureTelemetry(nodeID, dirPath, chainStacks, config.TelemetryPublisher)
+	if err != nil {
+		log.Fatal().Err(err).Msg("configuring telemetry")
 	}
```
jsign, line 229: In main, all the telemetry wiring is now expressed in these three lines.

cmd/api/main.go
```diff
@@ -279,16 +258,13 @@ func main() {
...
-		if config.TelemetryPublisher.Enabled {
-			metricsPublisher.Close()
-		}
 		if err := executorsDB.Close(); err != nil {
 			log.Error().Err(err).Msg("closing executors db")
 		}
+
+		if err := closeTelemetryModule(ctx); err != nil {
+			log.Error().Err(err).Msg("closing telemetry module")
+		}
```
jsign, line 267: The logic of `cli.HandleInterrupt(...)` is indirectly simplified.

cmd/api/main.go
```diff
@@ -419,3 +396,77 @@ func createChainIDStack(
 		},
 	}, nil
 }
+
+func configureTelemetry(
```
jsign, line 400: All the wiring of the telemetry stuff happens in this function.
A new person can simply jump to this function to understand how the wiring works.
Also, the parameters clearly show configurations and dependencies needed.

cmd/api/main.go
```diff
@@ -404,6 +380,7 @@ func createChainIDStack(
 	return chains.ChainStack{
 		Store:                 systemStore,
 		Registry:              registry,
+		EventProcessor:        ep,
```
jsign, line 383: `chains.ChainStack` now also has the `EventProcessor`, this is needed to allow an external module to ask the `EventProcessor`, which is the last processed block. More on that later in the PR.

cmd/api/main.go
```diff
@@ -419,3 +396,77 @@ func createChainIDStack(
...
+			return nil, fmt.Errorf("creating metrics http exporter: %s", err)
+		}
+		publishingInterval, err := time.ParseDuration(config.PublishingInterval)
+		if err != nil {
+			return nil, fmt.Errorf("parsing publishing interval: %s", err)
+		}
+		metricsPublisher = publisher.NewPublisher(metricsStore, exporter, nodeID, publishingInterval)
+		metricsPublisher.Start()
+	}
+	telemetry.SetMetricStore(metricsStore)
```
jsign, line 433: All this section is the same that we had in main.

cmd/api/main.go
```diff
@@ -419,3 +396,77 @@ func createChainIDStack(
...
+
+	// Collect binary information.
+	ctx, cls := context.WithTimeout(context.Background(), time.Second)
+	defer cls()
+	if err := telemetry.Collect(ctx, buildinfo.GetSummary()); err != nil {
+		return nil, fmt.Errorf("collect git summary: %s", err)
+	}
```
jsign, line 440: We do the Git binary information collection.

cmd/api/main.go
```diff
@@ -419,3 +396,77 @@ func createChainIDStack(
...
+	chainStackCollectorClosed := make(chan struct{})
+	cc, err := chainscollector.New(chainStacks, collectFrequency)
+	if err != nil {
+		clsChainStackCollector()
+		return nil, fmt.Errorf("configure chains collector: %s", err)
+	}
+	go func() {
+		cc.Start(ctxChainStackCollector)
+		close(chainStackCollectorClosed)
+	}()
```
jsign, line 456: We spin up a new telemetry sub-module `chainscollector`.
This is a background module that collects chains stacks information in metrics with a configured frequency.

cmd/api/main.go
```diff
@@ -419,3 +396,77 @@ func createChainIDStack(
...
+		<-chainStackCollectorClosed
+
+		if err := metricsStore.Close(); err != nil {
+			return fmt.Errorf("closing metrics store: %s", err)
+		}
+		if config.Enabled {
+			metricsPublisher.Close()
+		}
+
+		return nil
```
jsign, line 470: This is the closing function returned to `main`. This allows us to clearly express all the needed work to close the telemetry module properly. The logic here isn't mixed with other wires of main.

Also, `main` will simply have to execute this returned function without knowing what it does. It simply knows that it will do the right thing.

docker/deployed/staging/api/config.json
```diff
@@ -45,7 +45,8 @@
     "Enabled": true,
     "MetricsHubURL": "https://metricshub-staging-mrgr43cf5q-uw.a.run.app",
     "MetricsHubApiKey": "${METRICS_HUB_API_KEY}",
-    "PublishingInterval": "10s"
+    "PublishingInterval": "10s",
+    "ChainStackCollectFrequency": "15m"
```
jsign, line 49: 15m was an arbitrary value that I think is reasonable. I'm open to other values here. LMK.

pkg/eventprocessor/impl/eventprocessor.go
```diff
@@ -112,6 +112,16 @@ func (ep *EventProcessor) Start() error {
...
 
+// GetLastExecutedBlockNumber returns the last executed block number.
+func (ep *EventProcessor) GetLastExecutedBlockNumber(ctx context.Context) (int64, error) {
+	blockNumber, err := ep.executor.GetLastExecutedBlockNumber(ctx)
+	if err != nil {
+		return 0, fmt.Errorf("get last executed block number: %s", err)
+	}
+
+	return blockNumber, nil
+}
```
jsign, line 123: Now the `EventProcessor` can be asked which was the last executed block number.

pkg/telemetry/chainscollector/chainscollector.go
```diff
@@ -0,0 +1,65 @@
+package chainscollector
```
jsign, line 1: This is the new telemetry sub-module that collects chains information with some frequency.

pkg/telemetry/chainscollector/chainscollector.go
```diff
@@ -0,0 +1,65 @@
...
+			}
+		}
+	}
+}
+
+type chainIDBlockNumbers map[tableland.ChainID]int64
+
+func (cbn chainIDBlockNumbers) GetLastProcessedBlockNumber() map[tableland.ChainID]int64 {
+	return cbn
+}
```
jsign, line 65: `chainIDBLockNumbers` implements the new `telemetry.ChainsStackSummary` metric interface.

pkg/telemetry/metrics.go
```diff
@@ -5,6 +5,7 @@ import (
 	"time"
```
jsign, line 5: We add here the two new interfaces for the metrics, and their corresponding structs for deserialization.

pkg/telemetry/storage/db_test.go
```diff
@@ -10,19 +10,29 @@ import (
...
-		telemetry.SetMetricStore(s)
+	t.Parallel()
+
+	// Notes:
+	// This can't be wired per sub-tests for two reasons:
+	// 1- `telemetry.SetMetricStore(...)` is a global setup at the package level, and
+	// 2- `SetMetricStore(...)` has a `sync.Once` wrapping so can't be called more than once, so each sub-test can't
+	//     override their value.
+	//
+	// This also means that sub-tests can't run in parallel.
```
jsign, line 29: Before this PR, this test only had a single sub-test.
Adding extra sub-tests had some surprising effects, explained in this comment.

It is not a big deal, but we can't strictly spin up a "complete fresh stack" per sub-test for the explained reasons.

LMK if it makes sense.

pkg/telemetry/storage/db_test.go
```diff
@@ -10,19 +10,29 @@ import (
...
+	dbURI := tests.Sqlite3URI(t)
+	s, err := New(dbURI)
+	require.NoError(t, err)
+	telemetry.SetMetricStore(s)
 
+	t.Run("state hash", func(t *testing.T) {
```
jsign, line 35: This test was only simplified a bit. Unfortunately, the diff got mixed with the extra sub-tests.

pkg/telemetry/storage/db_test.go
```diff
@@ -31,17 +41,100 @@ func TestCollectAndFetchAndPublish(t *testing.T) {
...
+		require.Eventually(t, func() bool {
+			metrics, err = s.FetchUnpublishedMetrics(context.Background(), 2)
+			require.NoError(t, err)
+			return len(metrics) == 0
+		}, 5*time.Second, time.Second)
+
+		p.Close()
+	})
+
+	t.Run("git summary", func(t *testing.T) {
```
jsign, line 75: New test, nothing interesting.

pkg/telemetry/storage/db_test.go
```diff
@@ -31,17 +41,100 @@ func TestCollectAndFetchAndPublish(t *testing.T) {
...
+
+		require.Eventually(t, func() bool {
+			metrics, err = s.FetchUnpublishedMetrics(context.Background(), 2)
+			require.NoError(t, err)
+			return len(metrics) == 0
+		}, 5*time.Second, time.Second)
+
+		p.Close()
+	})
+	t.Run("chains stack summary", func(t *testing.T) {
```
jsign, line 117: New test, nothing interesting.

pkg/telemetry/telemetry.go
```diff
@@ -62,7 +62,38 @@ func Collect(ctx context.Context, metric interface{}) error {
...
+		if err := metricStore.StoreMetric(ctx, Metric{
+			Version:   1,
+			Timestamp: time.Now().UTC(),
+			Type:      ChainStacksSummaryType,
+			Payload: ChainStacksMetric{
+				Version:                   1,
+				LastProcessedBlockNumbers: v.GetLastProcessedBlockNumber(),
+			},
+		}); err != nil {
+			return errors.Errorf("store git summary metric: %s", err)
```
brunocalza, line 96: nit: `store chain stacks summary metric`

brunocalza (approved):
Looks good! made a comment about the frequency config. That value seems too close to the frequency we are collecting the state hash. If you think that's fine, no problem :)

cmd/api/main.go
brunocalza: nice

docker/deployed/testnet/api/config.json
```diff
@@ -46,7 +46,8 @@
         "Enabled": true,
         "MetricsHubURL": "https://metricshub-testnet-mrgr43cf5q-uw.a.run.app",
         "MetricsHubApiKey": "${METRICS_HUB_API_KEY}",
-        "PublishingInterval": "10s"
+        "PublishingInterval": "10s",
+        "ChainStackCollectFrequency": "15m"
```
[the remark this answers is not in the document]
jsign: The state-hash "frequency" is hand-wavy and not strictly constant for many reasons:
- The frequency is defined by "# blocks' and is a per-chain configuration, not global.
- The "30 mins" is an indirect result of the values we baked today, but I'm not sure how that will play out in the future.
- More fundamentally, remember that calculating the state hash of every X blocks (where X is today calculated to be ~30min with some estimations), is a hand-wavy frequency (not fixed) since also depends when a transaction happens in that chain. e.g: if a chain doesn't have a transaction in 80mins, then no state-hash will be calculated since that is triggered "on the next txn that happens after block %X=0".

That was my thought process. The last point might be the most annoying one, since we wouldn't have any metric with a fixed frequency to also infer uptime.

I think it would be totally fine to merge both, but we should be aware of this "irregular heartbeat".

pkg/telemetry/storage/db_test.go
brunocalza: yeah. makes sense. that's the trade-off of having those global variables

cmd/api/main.go
brunocalza: 👍 as a suggestion for later. we can move to an `App` struct that has methods for configuring the "app". And in the main, we would have something like `app.ConfigureXXX().ConfigureYYY().Run()`.

PR #343, opened 2022-10-12
otelhttp: add service name

jsign:
# Summary

This PR solves, hopefully, the last breaking change of OTEL default attributes behavioral change.

In the testnet dashboard, some tiles don't show information:
image

In the staging dashboard I deployed this PR and it shows data correctly.

# Context

https://github.com/tablelandnetwork/go-tableland/pull/323

# Implementation overview

Luckily, after digging into the `otelhttp`, I saw that you can inject custom labels if you do an extra middleware wrapping.
It'll become clear while seeing the code.

NA

### DIFF
--- a/internal/router/middlewares/otelhttp.go
+++ b/internal/router/middlewares/otelhttp.go
@@ -3,12 +3,23 @@ package middlewares
 import (
 	"net/http"
 
+	"github.com/textileio/go-tableland/pkg/metrics"
 	"go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp"
 )
 
 // OtelHTTP wraps the handler h with OTEL metrics.
 func OtelHTTP(operation string) func(h http.Handler) http.Handler {
 	return func(h http.Handler) http.Handler {
-		return otelhttp.NewHandler(h, operation)
+		return otelhttp.NewHandler(&labeledHandler{h: h}, operation)
 	}
+}
+
+type labeledHandler struct {
+	h http.Handler
+}
+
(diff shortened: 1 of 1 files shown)


internal/router/middlewares/otelhttp.go
```diff
@@ -3,12 +3,23 @@ package middlewares
...
+		return otelhttp.NewHandler(&labeledHandler{h: h}, operation)
 	}
 }
+
+type labeledHandler struct {
+	h http.Handler
+}
+
+func (lh *labeledHandler) ServeHTTP(rw http.ResponseWriter, r *http.Request) {
+	labeler, _ := otelhttp.LabelerFromContext(r.Context())
```
jsign, line 22: The `otelhttp.NewHandler` in L13 creates a `http.Handler` that internally creates an internal `&Labeler{}`.

That means that in wrapped internal `http.Handlers`, middlewares can do `otelhttp.LabelerFromContext(...)` can get that "base labeler". That allows each handler to add custom extra attributes to be included in captured metrics.

This is exactly what we need. What we do is always include the `service_name` form the new `metrics.BaseAttrs`.

The same mechanics might be useful if we want to include custom labels in different handlers.  For example, we could try including the `chain_id`, so we can have all the `otelhttp` metrics we have today, but also separated by chain_id, which we don't have.

We can think if that's useful in the future, it would be an easy addition. I'd have to just do both L22-L23-ish in the "chain id" detection middleware.

brunocalza: Nice!

PR #349, opened 2022-10-13
Removes system_id entries when restoring from backup

brunocalza:
# Summary

When restoring from backup we need to delete `system_id` entries

# Context

Node operators who bootstrapped a node from a backup file were producing metrics with the same node id as of of our validators and we could not distinguish the metrics. 

# Implementation overview

Simply removes the entry from the `system_id` file.

To be honest I don't think this is the best solution. Every time a node bootstraps from backup it will generate a new id. A better approach would be to keep the existing id but that solution requires a bit more work and thought. 

na

# Checklist


cmd/metricshub/main.go
```diff
@@ -92,6 +95,13 @@ func makeHandler(store store, c *config) http.HandlerFunc {
 			return
 		}
 
+		ip, err := extractClientIP(r)
+		if err != nil {
+			log.Warn().Msg("could not get ip address")
+		}
+
+		log.Debug().Str("node_id", req.NodeID).Str("ip_address", ip).Msg("got metric from node")
```
brunocalza, line 103: Added this to see if we are getting the same node id from different IP address in the metrics hub server

jsign (approved):

PR #371, opened 2022-10-27
Improves the backup restorer to consider an existing database

brunocalza:
# Summary

A small improvement in the backup restoration process to consider that a database could already exist and contain local information such as node id and pending transactions

# Context

The backup restorer was deleting the node id and pending txs from the backup database. But that turns out to be a bad idea. The node that is bootstrapping from a backup file may have pending txs in its local databases, and it already has a node id. Deleting those means that we lose track of the pending txs, and the node is forced to generate a new node id, which is not ideal.

# Implementation overview

- We check if a database already exists
- If a database already exists, rename it to another name
- Create a new database that is a copy of the backup
- Delete information from `system_pending_tx` and `system_id` tables
- Copy information of `system_pending_txs` and `system_id` tables from the existing renamed database to the new database
- Delete existing renamed database

na

# Checklist

### DIFF
--- a/pkg/backup/restorer/restorer.go
+++ b/pkg/backup/restorer/restorer.go
@@ -2,6 +2,7 @@ package restorer
 
 import (
 	"database/sql"
+	"errors"
 	"fmt"
 	"io"
 	"net/http"
@@ -101,8 +102,16 @@ func (br *BackupRestorer) load() error {
 		}
 	}()
 
-	if err := os.Remove(fmt.Sprintf("%s/%s", filepath.Dir(br.dbPath), filepath.Base(br.dbPath))); err != nil {
-		log.Warn().Err(err).Msg("removing database file")
+	dbPath := fmt.Sprintf("%s/%s", filepath.Dir(br.dbPath), filepath.Base(br.dbPath))
+	tmpDbPath := dbPath + ".tmp"
+
+	// If a database already exists we are going to rename it,
+	// so, later, we can get the node id information from the existing database.
+	dbExists := fileExists(dbPath)
+	if dbExists {
+		if err := os.Rename(dbPath, tmpDbPath); err != nil {
+			log.Warn().Err(err).Msg("renaming file")
+		}
 	}
 
 	if err := os.Remove(fmt.Sprintf("%s/%s-wal", filepath.Dir(br.dbPath), filepath.Base(br.dbPath))); err != nil {
@@ -113,7 +122,7 @@ func (br *BackupRestorer) load() error {
 		log.Warn().Err(err).Msg("removing database shm file")
 	}
 
-	out, err := os.Create(fmt.Sprintf("%s/%s", filepath.Dir(br.dbPath), filepath.Base(br.dbPath)))
+	out, err := os.Create(dbPath)
 	if err != nil {
 		return fmt.Errorf("creating file: %s", err)
 	}
@@ -128,7 +137,7 @@ func (br *BackupRestorer) load() error {
 		return fmt.Errorf("copying file: %s", err)
 	}
 
-	db, err := sql.Open("sqlite3", fmt.Sprintf("%s/%s", filepath.Dir(br.dbPath), filepath.Base(br.dbPath)))
+	db, err := sql.Open("sqlite3", dbPath)
 	if err != nil {
 		return fmt.Errorf("opening database: %s", err)
 	}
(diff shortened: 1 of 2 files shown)


pkg/backup/restorer/restorer_test.go
```diff
@@ -13,7 +13,7 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-func TestRestorer(t *testing.T) {
+func TestRestorerWithNoExistingDatabase(t *testing.T) {
```
brunocalza, line 16: Renamed this to indicate this is a test with no existing database

pkg/backup/restorer/restorer_test.go
```diff
@@ -54,3 +54,85 @@ func TestRestorer(t *testing.T) {
 	require.NoError(t, err)
 	require.Equal(t, 0, c)
 }
+
+func TestRestorerWithExistingDatabase(t *testing.T) {
```
brunocalza, line 58: This is a new test with an existing database

pkg/backup/restorer/restorer_test.go
```diff
@@ -54,3 +54,85 @@ func TestRestorer(t *testing.T) {
...
+	require.NoError(t, err)
+
+	db, err := sql.Open("sqlite3", databaseURL)
+	require.NoError(t, err)
+
+	var a int
+	err = db.QueryRow("SELECT a FROM a LIMIT 1").Scan(&a)
+	require.NoError(t, err)
+	require.Equal(t, 1, a)
+
```
brunocalza, line 93: We need to check that those tables contain the database from the existing table, and are not empty as in the other test

pkg/backup/restorer/restorer_test.go
```diff
@@ -54,3 +54,85 @@ func TestRestorer(t *testing.T) {
...
+	require.NoError(t, err)
+	require.Equal(t, "existing address", address)
+
+	var nodeID string
+	err = db.QueryRow("SELECT id FROM system_id").Scan(&nodeID)
+	require.NoError(t, err)
+	require.Equal(t, "existing node id", nodeID)
+}
+
+func createExistingDatabase(t *testing.T, dir string) {
```
brunocalza, line 105: Create a database with `system_id` and `system_pending_tx` tables with one row each

pkg/backup/restorer/restorer.go
```diff
@@ -128,22 +137,37 @@ func (br *BackupRestorer) load() error {
...
+	db, err := sql.Open("sqlite3", dbPath)
 	if err != nil {
 		return fmt.Errorf("opening database: %s", err)
 	}
 	defer func() {
 		if err := db.Close(); err != nil {
 			log.Error().Err(err).Msg("closing db")
 		}
+
+		if dbExists {
```
brunocalza, line 149: we remove the temporary db if it exists after the new database is closed

pkg/backup/restorer/restorer.go
```diff
@@ -101,8 +102,16 @@ func (br *BackupRestorer) load() error {
 		}
 	}()
 
-	if err := os.Remove(fmt.Sprintf("%s/%s", filepath.Dir(br.dbPath), filepath.Base(br.dbPath))); err != nil {
-		log.Warn().Err(err).Msg("removing database file")
+	dbPath := fmt.Sprintf("%s/%s", filepath.Dir(br.dbPath), filepath.Base(br.dbPath))
+	tmpDbPath := dbPath + ".tmp"
```
jsign, line 106: Nit: can we rename `tmpDbPath` to `previousDbPath`?

pkg/backup/restorer/restorer.go
```diff
@@ -101,8 +102,16 @@ func (br *BackupRestorer) load() error {
...
+
+	// If a database already exists we are going to rename it,
+	// so, later, we can get the node id information from the existing database.
+	dbExists := fileExists(dbPath)
+	if dbExists {
+		if err := os.Rename(dbPath, tmpDbPath); err != nil {
+			log.Warn().Err(err).Msg("renaming file")
+		}
```
jsign, line 114: Maybe it's safer to return an error and let the operator figure out why this failed.
If the renaming didn't work correctly, I've a sense something somewhat bad could happen if we continue.

pkg/backup/restorer/restorer.go
```diff
@@ -128,22 +137,37 @@ func (br *BackupRestorer) load() error {
...
 		return fmt.Errorf("opening database: %s", err)
 	}
 	defer func() {
 		if err := db.Close(); err != nil {
 			log.Error().Err(err).Msg("closing db")
 		}
+
+		if dbExists {
+			if err := os.Remove(tmpDbPath); err != nil {
+				log.Error().Err(err).Msg("closing db")
```
jsign, line 151: Nit: change message to `removing previous db` or similar.

pkg/backup/restorer/restorer.go
```diff
@@ -160,3 +184,16 @@ func (br *BackupRestorer) cleanUp() error {
...
+func fileExists(name string) bool {
+	_, err := os.Stat(name)
+	if err == nil {
+		return true
+	}
+	if errors.Is(err, os.ErrNotExist) {
+		return false
+	}
+
+	log.Warn().Err(err).Msg("file exists")
```
jsign, line 197: Nit: maybe add `.Str("file_name", name)` to the chain to make this log line context more understandable.

jsign (approved):
LGTM! Left some comments for your consideration.

PR #379, opened 2022-10-31
Fix #378 allow config dir path as command flag

### LINKED ISSUE #378: How to use `--dir` flag?
I've been successfully building and running this repo with the goal of finishing https://github.com/orgs/tablelandnetwork/projects/36
But passing in the location of the config.json file with the `--dir` flag always results in `flag provided but not defined: -dir`

### Steps to reproduce

1. build with `BIN_BUILD_FLAGS="GOOS=darwin" make build-api-debug`.  I also tried `BIN_BUILD_FLAGS="GOOS=linux" make build-api-debug`
2. run the output of the build with `./api --dir <non-default path to config>`

I'm working on a intel Mac with Big Sur 11.5.2
go version go1.19.1 darwin/amd64
make --version
GNU Make 3.81
Copyright (C) 2006  Free Software Foundation, Inc.
This is free software; see the source for copying conditions.
There is NO warranty; not even for MERCHANTABILITY or FITNESS FOR A
PARTICULAR PURPOSE.

This program built for i386-apple-darwin11.3.0

### What I've tried

If I change the default value of the dir flag to the path of my local machine and do *not* include the flag when running, everything works fine.  However my current thinking would require the path to the config is specific to the machine running the validator.

I searched for similar issues and found a suggestion `flag.Parse()` is being called before the flag is defined, but the only useage of the flag package I can find is the origin of the error: https://github.com/tablelandnetwork/go-tableland/blob/main/cmd/api/config.go#L114-L140

I suspect that the problem is either me making a simple mistake, or something to do with the go uconfig package.
Any help would be greatly appreciated

joewagner:
TL;DR; uconfig enforces that all flags be defined in the config struct
See the discussion in #378 for more detail.

### DIFF
--- a/cmd/api/config.go
+++ b/cmd/api/config.go
@@ -18,6 +18,7 @@ import (
 var configFilename = "config.json"
 
 type config struct {
+	Dir string
 	HTTP struct {
 		Port string `default:"8080"` // HTTP port (e.g. 8080)
 
--- a/cmd/api/config.go
+++ b/cmd/api/config.go
@@ -18,7 +18,7 @@ import (
 var configFilename = "config.json"
 
 type config struct {
-	Dir string
+	Dir  string
 	HTTP struct {
 		Port string `default:"8080"` // HTTP port (e.g. 8080)
 
--- a/cmd/api/config.go
+++ b/cmd/api/config.go
@@ -18,7 +18,7 @@ import (
 var configFilename = "config.json"
 
 type config struct {
-	Dir  string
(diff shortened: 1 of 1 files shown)


brunocalza (approved):
I think this is good enough to get you unstuck. 

There is a small implication though. When we access `config.Dir` later, and the flag was not set, it will have an empty string value, instead of the default. Since we are not relying on `config.Dir` for anything, that's fine. Maybe worth adding a comment to indicate that field should not be used.

PR #381, opened 2022-11-01
main: refactor wirings

### LINKED ISSUE #302: Consider re-organizing dependencies creation

jsign:
# Summary

This PR does a general refactor in `main.go` to organize our wires better. The main goal is to have a `func main()` that's more understandable.

# Context

Closes #302 

# Implementation overview

The `main()` function is now more high-level, always calling `create***()` functions that will create modules. These methods usually return a `close***` variable, a `moduleCloser` function to be used when Ctrl+C is received.

To clarify things, I've extracted anonymous structs used in `config.go` to be non-anonymous.

NA

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -36,7 +36,6 @@ linters-settings:
 
 linters:
   enable:
-    - revive
     - misspell
     - bodyclose
     - unconvert
--- a/cmd/api/config.go
+++ b/cmd/api/config.go
@@ -18,46 +18,59 @@ import (
 var configFilename = "config.json"
 
 type config struct {
-	Dir  string // This will default to "", NOT the default dir value set via the flag package
-	HTTP struct {
-		Port string `default:"8080"` // HTTP port (e.g. 8080)
+	Dir                string // This will default to "", NOT the default dir value set via the flag package
+	BootstrapBackupURL string `default:"" env:"BOOTSTRAP_BACKUP_URL"`
 
-		TLSCert string `default:""`
-		TLSKey  string `default:""`
+	HTTP    HTTPConfig
+	Gateway GatewayConfig
 
-		RateLimInterval       string `default:"1s"`
-		MaxRequestPerInterval uint64 `default:"10"`
-	}
-	Gateway struct {
-		ExternalURIPrefix   string `default:"https://testnet.tableland.network"`
-		MetadataRendererURI string `default:""`
-	}
 	TableConstraints TableConstraints
 	QueryConstraints QueryConstraints
-	Metrics          struct {
+
+	Metrics struct {
 		Port string `default:"9090"`
 	}
 	Log struct {
 		Human bool `default:"false"`
 		Debug bool `default:"false"`
 	}
-	Chains    []ChainConfig
 	Analytics struct {
 		FetchExtraBlockInfo bool `default:"false"`
 	}
-	Backup struct {
-		Enabled           bool   `default:"true"`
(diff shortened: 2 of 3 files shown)


.golangci.yml
```diff
@@ -36,7 +36,6 @@ linters-settings:
 
 linters:
   enable:
-    - revive
```
jsign, line 39: I'm removing `revive` since, for some reason, it was firing that `moduleCloser`, which is `func(context.Context) error`, is a _duplicate_ of an existing type (e.g: `io.Close`). 

OK, that's true, but I won't use `io.Closer`. In any case, I'm not sure I've ever seen any useful lint from `revive` anyway.

cmd/api/config.go
```diff
@@ -18,46 +18,59 @@ import (
 var configFilename = "config.json"
```
jsign, line 18: TL;DR de-anonymizing config structs, so we can pass more focused parameters to `main` method builders.

cmd/api/main.go
```diff
@@ -47,241 +47,106 @@ import (
 	"go.opentelemetry.io/otel/attribute"
```
jsign, line 47: I'm not sure if looking at diffs here is fine, or it's better to see the real file.

cmd/api/main.go
```diff
@@ -47,241 +47,106 @@ import (
...
-	log.Info().Str("node_id", nodeID).Msg("node info")
-
-	closeTelemetryModule, err := configureTelemetry(nodeID, dirPath, chainStacks, config.TelemetryPublisher)
+	// Telemetry
+	closeTelemetryModule, err := configureTelemetry(dirPath, chainStacks, config.TelemetryPublisher)
 	if err != nil {
 		log.Fatal().Err(err).Msg("configuring telemetry")
 	}
 
 	cli.HandleInterrupt(func() {
```
jsign, line 122: Hopefully, this "closing logic" is clearer now.

cmd/api/main.go
```diff
@@ -490,3 +365,215 @@ func configureTelemetry(
 		return nil
 	}, nil
 }
+
+func restoreBackup(databaseURL string, backupURL string) error {
```
jsign, line 369: Each new builder function tries to follow to general rules:
1) Receive as parameters the most concrete type as possible.
2) Return a `closerModule` function if whatever it built needs to be gracefully closed by main.

cmd/api/main.go
```diff
@@ -490,3 +365,215 @@ func configureTelemetry(
...
+	log.Info().Msg("starting backup restore")
+	elapsedTime := time.Now()
+	if err := restorer.Restore(); err != nil {
+		return fmt.Errorf("backup restoration failed: %s", err)
+	}
+	log.Info().Float64("elapsed_time_seconds", time.Since(elapsedTime).Seconds()).Msg("backup restore finished")
+	return nil
+}
+
+func createParser(queryConstraints QueryConstraints) (parsing.SQLValidator, error) {
```
jsign, line 384: An example of sending `QueryConstraints`, and not the full `Config`.
Being focused allows to truly understand what configs or dependencies are minimally needed in different modules.

brunocalza (approved):
Looks good to me 👍👍

PR #382, opened 2022-11-02
Adds sqlc back, with sqlite engine

### LINKED ISSUE #304: Switch to use sqlc again
When we switched from Postgres to SQLite, we had to stop using the great library https://github.com/kyleconroy/sqlc, because then it didn't have support for SQLite. Since then, we've been evolving the code in a similar way with the same folder structure.

Some time ago was announced that SQLite is now supported, so we can (and I think, should) start using this library again. It makes maintaining database access for a big part of the validator much easier and less error-prone.

[See here for docs](https://docs.sqlc.dev/en/latest/tutorials/getting-started-sqlite.html?highlight=sqlite).

brunocalza:
# Summary

Adds `sqlc` back now that it has support for SQLite. 

# Context

Closes #304 

# Implementation overview

NA

NA

# Checklist


Makefile
```diff
@@ -31,7 +32,7 @@ ethereum-testerc721a:
 .PHONY: ethereum-testerc721a
 
 system-sql-assets:
-	cd pkg/sqlstore/impl/system && $(GO_BINDATA) -pkg migrations -prefix migrations/ -o migrations/migrations.go -ignore=migrations.go migrations
+	cd pkg/sqlstore/impl/system && $(GO_BINDATA) -pkg migrations -prefix migrations/ -o migrations/migrations.go -ignore=migrations.go migrations && $(SQLC) generate; cd -;
```
brunocalza, line 35: Adds `sqlc` command back

pkg/sqlstore/impl/system/sqlc.yaml
```diff
@@ -0,0 +1,29 @@
+version: "2"
```
brunocalza, line 1: Using version 2 now

pkg/sqlstore/impl/system/sqlc.yaml
```diff
@@ -0,0 +1,29 @@
...
+sql:
+  - schema:
```
brunocalza, line 3: This is a bit annoying. Could not find another solution. One of our queries touches the `sqlite_master` table and that's an internal table that has no schema definition inside `migrations` and `sqlc` needs a schema definition. 

We could add a `CREATE TABLE IF NOT EXISTS sqlite_master` inside `migrations`. That solves the `sqlc` problem but that fails when running the migration because `sqlite_master` is internal. 

So, the solution I found was to create another directory called `schemas` to [hold schemas](pkg/sqlstore/impl/system/schemas/schemas.sql) that will not be run by migration.

pkg/sqlstore/impl/system/sqlc.yaml
```diff
@@ -0,0 +1,29 @@
...
+        package: "db"
+        out: "./internal/db"
+        emit_prepared_queries: true
+        emit_interface: false
+        emit_exact_table_names: false
+        emit_empty_slices: false
+        emit_json_tags: false
+        json_tags_case_style: "camel"
+        sql_package: "database/sql"
+        overrides:
```
brunocalza, line 19: I've added some type overrides to avoid a lot of casting inside [store.go](pkg/sqlstore/impl/system/store.go).

pkg/sqlstore/impl/system/store.go
```diff
@@ -97,8 +97,8 @@ func (s *SystemStore) GetTablesByController(ctx context.Context, controller stri
 		return []sqlstore.Table{}, fmt.Errorf("sanitizing address: %s", err)
 	}
```
brunocalza, line 98: Most of the changes in this file are to adjust for changes in the name of fields and their types. The names and types `sqlc` generated was not 100% compatible with our code. 

I played a lot with `sqlc` trying to generate the names and types compatible with existing code. But since our code evolved without `sqlc` we don't follow a consistent type schema for some fields. For example, block number is `int64` in some places and `uint64` in others. No matter what rule I put in `sqlc` in some place we have to do the casting.  The same happens with timestamp fields: `time.Time` and `int64`.

pkg/sqlstore/impl/system/store.go
```diff
@@ -401,11 +402,17 @@ func (s *SystemStore) SaveEVMEvents(ctx context.Context, events []tableland.EVME
 // It receives an optional fromHeight to only look for blocks after a block number. If null it will look
 // for blocks at any height.
 func (s *SystemStore) GetBlocksMissingExtraInfo(ctx context.Context, lastKnownHeight *int64) ([]int64, error) {
-	params := db.GetBlocksMissingExtraInfoParams{
-		ChainID:    int64(s.chainID),
-		FromHeight: lastKnownHeight,
+	var blockNumbers []int64
```
brunocalza, line 405: I made a change here because `sqlc` was not able to generate the old query correctly:
```
SELECT DISTINCT e.block_number
FROM system_evm_events e 
WHERE e.chain_id=?1 AND 
      (?2 is null OR e.block_number>?2) AND
      NOT EXISTS(select* from system_evm_blocks b WHERE e.chain_id=b.chain_id AND e.block_number=b.block_number)
ORDER BY e.block_number
```

[I broke that query in two](https://github.com/tablelandnetwork/go-tableland/pull/382/files#diff-8b677e635e3f5c4e0e23b5723035b61bc814cd63074425f673d7d941e51958fdR11-R22) and depending if the `lastKnownHeight` is set or not I pick the right query.

pkg/sqlstore/impl/system/store.go
jsign: :+1:

pkg/sqlstore/impl/system/sqlc.yaml
jsign: Nice, good solution.

jsign (approved):
Great! Very nice to have `sqlc` back :fire:

PR #385, opened 2022-11-04
Collects read query metrics 

### LINKED ISSUE #319: Gather information about read-queries
Today we collect operational (Prometheus) metrics about read-queries, histogram latencies, and similar. We show this in our operational dashboard, which is quite useful.

But I feel we're missing something extra: visibility in outliers. 
Once in a while, I see spikes in the 95-th latency in the `/query` endpoint, which is the result of sometimes a user sending complex/big queries.

It would be useful to know:
- Which are these queries?
- Who is sending these queries, and see if it's worth taking some appropriate action of "banning" per user/IP-address.
- Looking again at our timeouts for read-query execution to put some hard bounds.

Additionally, it might be interesting to simply collect all the read-queries (or do some sampling, but I don't think the volume is that big) so we can understand better how people are using the network to improve their experience or have better focus points. This can be an optional feature to enable in the validator since it has some storage overhead.

cc @sanderpick @brunocalza @carsonfarmer @dtbuchholz

brunocalza:
# Summary

Starts collecting read query data. Every time we receive a read query, either from REST or JSON RPC, we save the SQL statement, the IP Address, and the formatting options as a metric that will be shipped to BigQuery to be analyzed. 

# Context

Close #319 

# Implementation overview

Most of the implementation is creating the metric and adjusting things so the metric can be collected. 
Other than that, we are now deleting metrics older than 7 days to avoid storage problems. 

NA

### DIFF
--- a/internal/router/controllers/user.go
+++ b/internal/router/controllers/user.go
@@ -350,7 +350,7 @@ func CollectReadQueryMetric(ctx context.Context, statement string, config format
 			Output:  string(config.Output),
 		}
 
-		metric := &ReadQueryMetricData{
+		metric := &readQueryMetricData{
 			ipAddress:     ipAddress,
 			sqlStatement:  statement,
 			formatOptions: formatOptions,
@@ -364,16 +364,16 @@ func CollectReadQueryMetric(ctx context.Context, statement string, config format
 	}
 }
 
-type ReadQueryMetricData struct {
+type readQueryMetricData struct {
 	ipAddress     string
 	sqlStatement  string
 	formatOptions telemetry.ReadQueryFormatOptions
 	tookMilli     int64
 }
 
-func (m *ReadQueryMetricData) IPAddress() string    { return m.ipAddress }
-func (m *ReadQueryMetricData) SQLStatement() string { return m.sqlStatement }
-func (m *ReadQueryMetricData) FormatOptions() telemetry.ReadQueryFormatOptions {
+func (m *readQueryMetricData) IPAddress() string    { return m.ipAddress }
+func (m *readQueryMetricData) SQLStatement() string { return m.sqlStatement }
+func (m *readQueryMetricData) FormatOptions() telemetry.ReadQueryFormatOptions {
 	return m.formatOptions
 }
-func (m *ReadQueryMetricData) TookMilli() int64 { return m.tookMilli }
+func (m *readQueryMetricData) TookMilli() int64 { return m.tookMilli }
--- a/internal/router/controllers/user.go
+++ b/internal/router/controllers/user.go
@@ -11,9 +11,11 @@ import (
 	"github.com/hetiansu5/urlquery"
 	"github.com/rs/zerolog/log"
 	"github.com/textileio/go-tableland/internal/formatter"
+	"github.com/textileio/go-tableland/internal/router/middlewares"
 	"github.com/textileio/go-tableland/internal/tableland"
 	"github.com/textileio/go-tableland/pkg/errors"
 	"github.com/textileio/go-tableland/pkg/tables"
+	"github.com/textileio/go-tableland/pkg/telemetry"
 )
 
 // SQLRunner defines the run SQL interface of Tableland.
@@ -234,8 +236,9 @@ func (c *UserController) GetTableQuery(rw http.ResponseWriter, r *http.Request) 
 		return
 	}
 
-	rw.WriteHeader(http.StatusOK)
+	c.collectReadQueryMetric(r, stm, config)
 
+	rw.WriteHeader(http.StatusOK)
 	if config.Unwrap && len(res.Rows) > 1 {
 		rw.Header().Set("Content-Type", "application/jsonl+json")
 	}
@@ -265,6 +268,23 @@ func (c *UserController) runReadRequest(
 		return nil, false
 	}
 	return res, true
+}
+
+func (c *UserController) collectReadQueryMetric(r *http.Request, statement string, config formatter.FormatConfig) {
+	value := r.Context().Value(middlewares.ContextIPAddress)
+	ipAddress, ok := value.(string)
+	if ok && ipAddress != "" {
+		metric := &ReadQueryMetricData{
+			ipAddress:    ipAddress,
+			sqlStatement: statement,
+			extract:      config.Extract,
+			unwrap:       config.Unwrap,
+			output:       string(config.Output),
+		}
+		if err := telemetry.Collect(r.Context(), metric); err != nil {
+			log.Warn().Err(err).Msg("failed to collect metric")
+		}
+	}
 }
 
 func formatterOptions(r *http.Request) ([]formatter.FormatOption, error) {
@@ -331,3 +351,17 @@ func getFormatterParams(r *http.Request) (formatterParams, error) {
 
 	return c, nil
 }
+
+type ReadQueryMetricData struct {
+	ipAddress    string
+	sqlStatement string
+	output       string
+	extract      bool
+	unwrap       bool
+}
+
+func (m ReadQueryMetricData) IPAddress() string    { return m.ipAddress }
+func (m ReadQueryMetricData) SQLStatement() string { return m.sqlStatement }
+func (m ReadQueryMetricData) Output() string       { return m.output }
+func (m ReadQueryMetricData) Extract() bool        { return m.extract }
+func (m ReadQueryMetricData) Unwrap() bool         { return m.unwrap }
--- a/internal/router/controllers/user.go
+++ b/internal/router/controllers/user.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"net/http"
 	"strconv"
+	"time"
 
 	"github.com/gorilla/mux"
 	"github.com/hetiansu5/urlquery"
@@ -214,10 +215,12 @@ func (c *UserController) GetTableQuery(rw http.ResponseWriter, r *http.Request) 
 	rw.Header().Set("Content-Type", "application/json")
 
 	stm := r.URL.Query().Get("s")
+	start := time.Now()
 	res, ok := c.runReadRequest(r.Context(), stm, rw)
 	if !ok {
 		return
 	}
+	took := time.Since(start)
 
 	opts, err := formatterOptions(r)
 	if err != nil {
@@ -236,7 +239,7 @@ func (c *UserController) GetTableQuery(rw http.ResponseWriter, r *http.Request) 
 		return
 	}
 
-	c.collectReadQueryMetric(r, stm, config)
+	CollectReadQueryMetric(r.Context(), stm, config, took)
 
 	rw.WriteHeader(http.StatusOK)
 	if config.Unwrap && len(res.Rows) > 1 {
@@ -268,23 +271,6 @@ func (c *UserController) runReadRequest(
 		return nil, false
 	}
 	return res, true
-}
-
-func (c *UserController) collectReadQueryMetric(r *http.Request, statement string, config formatter.FormatConfig) {
-	value := r.Context().Value(middlewares.ContextIPAddress)
-	ipAddress, ok := value.(string)
-	if ok && ipAddress != "" {
-		metric := &ReadQueryMetricData{
-			ipAddress:    ipAddress,
-			sqlStatement: statement,
-			extract:      config.Extract,
-			unwrap:       config.Unwrap,
-			output:       string(config.Output),
-		}
-		if err := telemetry.Collect(r.Context(), metric); err != nil {
-			log.Warn().Err(err).Msg("failed to collect metric")
-		}
-	}
 }
 
 func formatterOptions(r *http.Request) ([]formatter.FormatOption, error) {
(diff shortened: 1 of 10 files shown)


internal/router/controllers/user.go
```diff
@@ -234,8 +236,9 @@ func (c *UserController) GetTableQuery(rw http.ResponseWriter, r *http.Request)
 		return
 	}
 
-	rw.WriteHeader(http.StatusOK)
+	c.collectReadQueryMetric(r, stm, config)
```
brunocalza, line 239: Here's the first point that the read query metric is collected. The second point is [in the JSON RPC service](https://github.com/tablelandnetwork/go-tableland/pull/385/files#diff-62ca30c3e73e3613d5dacf5ca84bd0f83f386a0a2ae949142dc36322483bcc92R212).

Some observations:
- Only successful queries are being collected. I thought that collecting unsuccessful queries could be a bit spammy.
- I'm also collecting the format options. That way we can analyze if people are using those options
- The IP address is also being collected. It is captured by the [logging middleware](https://github.com/tablelandnetwork/go-tableland/pull/385/files#diff-8874654c7b1682c125a6e8dada089eb7b278ba1130a3687563b0f0a0baccbc49R13-R20) and passed via context.

internal/router/controllers/user.go
```diff
@@ -267,6 +270,23 @@ func (c *UserController) runReadRequest(
 	return res, true
 }
 
+func (c *UserController) collectReadQueryMetric(r *http.Request, statement string, config formatter.FormatConfig) {
+	value := r.Context().Value(middlewares.ContextIPAddress)
+	ipAddress, ok := value.(string)
+	if ok && ipAddress != "" {
```
brunocalza, line 276: Only collecting if IP address was detected. We can change this, to always collect if that's make more sense

internal/router/controllers/user.go
```diff
@@ -331,3 +351,17 @@ func getFormatterParams(r *http.Request) (formatterParams, error) {
 
 	return c, nil
 }
+
```
brunocalza, line 354: I'm not sure if I'm liking this design choice of having to provide an implementation of an interface defined in the telemetry package. Kind of a pain to have to define a new struct just for that. If we could define an anonymous implementation of an interface it would be better. Just mentioning if other ideas come to mind.

internal/router/middlewares/logging.go
```diff
@@ -9,17 +10,20 @@ import (
 // WithLogging logs requests and responses that contain useful information.
 func WithLogging(h http.Handler) http.Handler {
 	handler := func(rw http.ResponseWriter, r *http.Request) {
+		clientIP, err := extractClientIP(r)
```
brunocalza, line 13: Decided to use the logging middleware to capture the IP address for the following reasons:
- it already had the extract IP logic (but it was used only in case of error)
- it is used for both REST and RPC
- I did not want to create one more middleware just for that

internal/router/rpcservice/rpcservice.go
```diff
@@ -207,6 +209,8 @@ func (rs *RPCService) RunReadQuery(
 		return RunReadQueryResponse{}, errors.New("unwrapped results with more than one row aren't supported in JSON RPC API")
 	}
 
+	rs.collectReadQueryMetric(ctx, req.Statement, config)
```
brunocalza, line 212: Second point in which the metric is collected

pkg/telemetry/publisher/publisher.go
```diff
@@ -92,13 +92,19 @@ func (p *Publisher) publish(ctx context.Context) error {
...
+	if err := p.store.DeleteOlderThan(ctx, sevenDays); err != nil {
+		return fmt.Errorf("mark as published: %s", err)
+	}
+
 	return nil
 }
 
 // MetricsStore defines the API for fetching metrics and marking them as published.
 type MetricsStore interface {
-	FetchUnpublishedMetrics(context.Context, int) ([]telemetry.Metric, error)
```
brunocalza, line 100: This is an API change I made to make it possible to fetch unpublished or published metrics. That was because I needed this kind of flexibility in the tests.

pkg/telemetry/publisher/publisher.go
```diff
@@ -70,7 +70,7 @@ func (p *Publisher) Close() {
 }
 
 func (p *Publisher) publish(ctx context.Context) error {
-	metrics, err := p.store.FetchUnpublishedMetrics(ctx, p.fetchAmount)
+	metrics, err := p.store.FetchMetrics(ctx, false, p.fetchAmount)
```
brunocalza, line 73: [see](https://github.com/tablelandnetwork/go-tableland/pull/385/files#diff-0462fbe75bcf5eb36a891b36056b09a0741a93e96112f8bb753e59e255f2f3ffR105)

pkg/telemetry/publisher/publisher.go
```diff
@@ -92,13 +92,19 @@ func (p *Publisher) publish(ctx context.Context) error {
 		return fmt.Errorf("mark as published: %s", err)
 	}
 
+	sevenDays := 24 * 7 * time.Hour
```
brunocalza, line 95: Added the logic to delete metrics older than 7 days to avoid the metrics database getting too large. Is 7 days a good number?

pkg/telemetry/publisher/publisher_test.go
```diff
@@ -29,15 +29,16 @@ func TestPublisher(t *testing.T) {
 	p.Start()
 
 	require.Eventually(t, func() bool {
-		return store.Len() == 0
+		return store.Len() == 0 && store.deleteOlderThanCalled
```
brunocalza, line 32: Added this testing condition to make sure the method was called

pkg/telemetry/storage/db_test.go
```diff
@@ -148,13 +149,97 @@ func TestCollectAndFetchAndPublish(t *testing.T) {
...
+			metrics, err = s.FetchMetrics(context.Background(), false, 2)
 			require.NoError(t, err)
 			return len(metrics) == 0
 		}, 5*time.Second, time.Second)
 
 		p.Close()
 	})
+
+	t.Run("delete old metrics", func(t *testing.T) {
+		// clear store
```
brunocalza, line 204: Also, kind of regretting to have made store a global variable in the pkg. It makes things a lot easier for clients but a lot harder to test. I have to clear the store so this test runs correctly because the store has data from previous tests.

internal/router/controllers/user.go
brunocalza: Realizing now that maybe we could capture the time it took to run? wdyt?

jsign: > Realizing now that maybe we could capture the time it took to run? wdyt?

Yes, that sounds very useful.

internal/router/controllers/user.go
jsign: Sounds fine. I think this should always be the case. 
Maybe it's worth adding a warning log for the `else`?

internal/router/controllers/user.go
```diff
@@ -267,6 +270,23 @@ func (c *UserController) runReadRequest(
...
+func (c *UserController) collectReadQueryMetric(r *http.Request, statement string, config formatter.FormatConfig) {
+	value := r.Context().Value(middlewares.ContextIPAddress)
+	ipAddress, ok := value.(string)
+	if ok && ipAddress != "" {
+		metric := &ReadQueryMetricData{
+			ipAddress:    ipAddress,
+			sqlStatement: statement,
+			extract:      config.Extract,
+			unwrap:       config.Unwrap,
+			output:       string(config.Output),
```
jsign, line 282: This looks fine, but we could also compact all the config as a JSON. If in the future we add a new attribute to config, we'd have to remember to add a new field here.

No strong opinions, but might be more future-proof.

internal/router/controllers/user.go
jsign: Yes, it felt somewhat heavy the last time I added a metric. We can meditate around changing this.
Worth a GH issue to track?

internal/router/middlewares/logging.go
```diff
@@ -9,17 +10,20 @@ import (
...
+		if err != nil {
+			log.Warn().Err(err).Msg("can't extract client ip")
+			clientIP = "none"
```
jsign, line 16: Uhm, should we reconsider this existing `"none"` value to be aligned with [this](https://github.com/tablelandnetwork/go-tableland/pull/385/files#diff-5588b773fe67092e7366bf6608cb74cae8d1475902b773f5c18a87e305b50dc2R276)? Like, setting it `""` (empty)?

No strong opinions, but want to be sure your `!=""` intention wasn't `!="none"` (or switch `"none"` to `""`.)

internal/router/rpcservice/rpcservice.go
```diff
@@ -270,3 +274,34 @@ func (rs *RPCService) SetController(
...
+			sqlStatement: statement,
+			extract:      config.Extract,
+			unwrap:       config.Unwrap,
+			output:       string(config.Output),
+		}
+		if err := telemetry.Collect(ctx, metric); err != nil {
+			log.Warn().Err(err).Msg("failed to collect metric")
+		}
+	}
+}
```
jsign, line 293: Can we create a metric helper to avoid this [code duplication](https://github.com/tablelandnetwork/go-tableland/pull/385/files#diff-5588b773fe67092e7366bf6608cb74cae8d1475902b773f5c18a87e305b50dc2R273)?
This can also include the `ReadQueryMetricData` struct?

We can ignore this for a while since JSON-RPC will be gone in the medium term, but for a while both duplicates would exist. No strong opinions.

pkg/telemetry/publisher/publisher.go
jsign: SGTM

pkg/telemetry/storage/db.go
```diff
@@ -76,13 +76,14 @@ func (db *TelemetryDatabase) StoreMetric(ctx context.Context, metric telemetry.M
...
 	rows, err := db.sqlDB.QueryContext(ctx,
 		`SELECT rowid, version, timestamp, type, payload, published FROM system_metrics 
-		WHERE published is false 
+		WHERE published is ?1 
 		ORDER BY timestamp
-		LIMIT ?1`,
+		LIMIT ?2`,
+		published,
 		amount,
 	)
```
jsign, line 88: Q: Makes sense to move this and other raw sqls in this file to sqlc?

pkg/telemetry/storage/db_test.go
jsign: Yeah, I remember this pain when I wanted to parallelize the tests.

jsign (approved):
Great! :100: 

I left some comments for your consideration.

pkg/telemetry/storage/db.go
brunocalza: hum. this is a different database. I completely forgot about this when putting the `sqlc` back. I only put it in the main database. I think we should use `sqlc` here too. But probably do it in another ticket.

internal/router/rpcservice/rpcservice.go
brunocalza: Yeah. I'll think about this. Not sure where would be a good place to put it. After JSON-RPC deletion, it would stay in an awkward place, far from where it's used.

internal/router/middlewares/logging.go
brunocalza: Nice catch! Thanks

internal/router/controllers/user.go
brunocalza: We could 👍

internal/router/controllers/user.go
brunocalza: @jsign I've tried the JSON approach but ended up in a weird spot:

```
		formatOptions := struct {
			Extract bool   `json:"extract"`
			Unwrap  bool   `json:"unwrap"`
			Output  string `json:"output"`
		}{
			Extract: config.Extract,
			Unwrap:  config.Unwrap,
			Output:  string(config.Output),
		}

		metric := &ReadQueryMetricData{
			ipAddress:     ipAddress,
			sqlStatement:  statement,
			formatOptions: formatOptions,
			tookMilli:     took.Milliseconds(),
		}
```

- I could pass `config` directly to `ReadQueryMetricData` but that would generate a JSON out of the standard format, with first letters capitalized: `{"Extract":false,"Output":"objects","Unwrap":false}`. If want to make them compatible, I have to do this copying to another struct, and we are back to the original problem of possibly forgetting to add a new config.
- Also, the way the metric JSON is formatted is the responsibility of the metric defined at `metrics.go`, not really the client pushing the metric's data through the library. 

Not really sure how to solve this. I think I'll go back to the original implementation and assume that if the client wants to have the new config in the metric he has to add it manually.

jsign: I was thinking more about the client sending a `config_json` field that is a `string`. I'm not sure I'd be bothered by having capital letters in the field or similar. 

But switching back to "per field" as before sounds good to me too. I don't know how many of these format options will be exploited in the analytics side, so whatever feels better sounds good. If a new field becomes important, then probably that means we'll remember to add it to the metric so that it can guide us. We're already trying to guess what will be useful, so not worth extra work or guessing.

brunocalza: Or I can

```
		formatOptions := telemetry.ReadQueryFormatOptions{
			Extract: config.Extract,
			Unwrap:  config.Unwrap,
			Output:  string(config.Output),
		}

		metric := &ReadQueryMetricData{
			ipAddress:     ipAddress,
			sqlStatement:  statement,
			formatOptions: formatOptions,
			tookMilli:     took.Milliseconds(),
		}
```

Still have to add the field, but much better.

jsign: SGTM

jsign (approved):

PR #392, opened 2022-11-08
analytics: send detected Tableland events and chain blocks as telemetry data

### LINKED ISSUE #307: Move the analytics database and related infra to BigQuery
Following a suggestion by @brunocalza, we should have some ETL or decided process to move the analytics database information living in the SQLite validator to BigQuery.

This is much nicer since:
- Retool has a native interface against BigQuery, so we don't need to run `postlite`  in the validator VM to act as a reverse proxy to give a Postgres wire-protocol support for SQLite. Also, it potentially means removing firewall rules to let Retool access the VM.
- Analytics read load won't hit the validator, prioritizing syncing chains and serving read queries.
- We'd have easier access to analytics information for the team and community by directly accessing the BQ database to do ad-hoc queries.

jsign:
# Summary

Closes #307

# Context

Our current Retool dashboard is querying our cloud validator directly, which isn't convenient since:
- We're adding analytics load to our main validator.
- The way data is stored in SQLite isn't suitable for analytics.
- Since SQLite is a local database, we needed to add an extra infrastructure piece (postlite) to have a PostgreSQL wire format.

By sending the `system_evm_events` and `system_evm_blocks` as events, we can leverage saving this information in BigQuery and use it as the database for Retool. We'll continue to do other ad-hoc analytics apart from the main dashboard, so we can keep taking advantage of having access to this data to do heavy queries.

After this PR is running in `main` for some days (so we have some data), I'll change the Retool dashboard to start using the Big Query database.

# Implementation overview

We've concluded that we'd add the logic to send events and blocks as new metrics. The information is still saved locally in SQLite since this information is still useful to be queried locally and is part of our generated backups/snapshots. We might reevaluate in the future if this is the case. 

For example, this data isn't operational but accounts for >90% of the backup sizes, which isn't strictly needed for new validators spinup. At some point, we might upload two kinds of backups, one with full events and blocks information and the other with minimal information to bootstrap a node.

No specific orientation is needed.

# Checklist

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -59,6 +59,6 @@ jobs:
             # Wait to start and double-check we're running the expected version.
             # This helps having a green check in the GH Action be high-signal that everything is fine.
             sleep 5;
-            for i in 1 2 3 4; do [ $(curl --insecure --silent https://127.0.0.1/version | jq .GitCommit) = "\"${{ steps.prep.outputs.sha_short}}\"" ] && break || (if [ $i = 4 ]; then exit -1; else sleep 5; fi ) done 
+            for i in 1 2 3 4; do [ $(curl --insecure --silent https://127.0.0.1/version | jq .git_commit) = "\"${{ steps.prep.outputs.sha_short}}\"" ] && break || (if [ $i = 4 ]; then exit -1; else sleep 5; fi ) done 
             echo "All healthy!"
 
--- a/pkg/eventprocessor/eventfeed/impl/analytics.go
+++ b/pkg/eventprocessor/eventfeed/impl/analytics.go
@@ -4,6 +4,8 @@ import (
 	"context"
 	"math/big"
 	"time"
+
+	"github.com/textileio/go-tableland/pkg/telemetry"
 )
 
 var fetchBlockExtraInfoDelay = time.Second * 10
@@ -35,6 +37,19 @@ func (ef *EventFeed) fetchExtraBlockInfo(ctx context.Context) {
 						Err(err).
 						Int64("block_number", blockNumber).
 						Msg("get extra block info")
+					return
+				}
+				newBlockMetric := telemetry.NewBlockMetric{
+					Version:            telemetry.NewBlockMetricV1,
+					ChainID:            int(ef.chainID),
+					BlockNumber:        blockNumber,
+					BlockTimestampUnix: block.Time,
+				}
+				if err := telemetry.Collect(ctx, newBlockMetric); err != nil {
+					ef.log.Error().
+						Err(err).
+						Int64("block_number", blockNumber).
+						Msg("capturing new block metric")
 					return
 				}
 				if err := ef.systemStore.InsertBlockExtraInfo(ctx, blockNumber, block.Time); err != nil {
--- a/pkg/eventprocessor/eventfeed/impl/eventfeed.go
+++ b/pkg/eventprocessor/eventfeed/impl/eventfeed.go
@@ -22,6 +22,7 @@ import (
 	"github.com/textileio/go-tableland/pkg/eventprocessor/eventfeed"
 	"github.com/textileio/go-tableland/pkg/sqlstore"
 	tbleth "github.com/textileio/go-tableland/pkg/tables/impl/ethereum"
+	"github.com/textileio/go-tableland/pkg/telemetry"
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/metric/instrument/syncint64"
 	"go.uber.org/atomic"
@@ -374,7 +375,7 @@ func (ef *EventFeed) persistEvents(ctx context.Context, events []types.Log, pars
 
 	store := ef.systemStore.WithTx(tx)
 
-	txnsEventsExistInDB := map[common.Hash]bool{}
+	persistedTxnHashEvents := map[common.Hash]bool{}
 	tblEvents := make([]tableland.EVMEvent, 0, len(events))
 	for i, e := range events {
 		// If we already have registered events for the TxHash, we skip persisting this event.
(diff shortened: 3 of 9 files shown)


.github/workflows/deploy.yml
```diff
@@ -59,6 +59,6 @@ jobs:
             # Wait to start and double-check we're running the expected version.
             # This helps having a green check in the GH Action be high-signal that everything is fine.
             sleep 5;
-            for i in 1 2 3 4; do [ $(curl --insecure --silent https://127.0.0.1/version | jq .GitCommit) = "\"${{ steps.prep.outputs.sha_short}}\"" ] && break || (if [ $i = 4 ]; then exit -1; else sleep 5; fi ) done 
+            for i in 1 2 3 4; do [ $(curl --insecure --silent https://127.0.0.1/version | jq .git_commit) = "\"${{ steps.prep.outputs.sha_short}}\"" ] && break || (if [ $i = 4 ]; then exit -1; else sleep 5; fi ) done
```
jsign, line 62: Fixing the CD pipeline since now the used struct comes from the `telemetry` package, which has snake-cased tags.

pkg/eventprocessor/eventfeed/impl/analytics.go
```diff
@@ -37,6 +39,19 @@ func (ef *EventFeed) fetchExtraBlockInfo(ctx context.Context) {
...
+					BlockNumber:        blockNumber,
+					BlockTimestampUnix: block.Time,
+				}
+				if err := telemetry.Collect(ctx, newBlockMetric); err != nil {
+					ef.log.Error().
+						Err(err).
+						Int64("block_number", blockNumber).
+						Msg("capturing new block metric")
+					return
+				}
```
jsign, line 54: We save first the new block information as a metric, and _then_ in the database. This is important since we do it in reverse and save the metric failed, we would miss sending the metric again (i.e: since the way we detect if a block information has been fetched, is by checking if it was inserted in `system_evm_blocks`).

In case that `InsertBlockExtraInfo` fails, this block will be retried and duplicate the metric persistence, but this is fine since in this case we'll dedupe at the analytics layer.

pkg/eventprocessor/eventfeed/impl/eventfeed.go
```diff
@@ -374,7 +375,7 @@ func (ef *EventFeed) persistEvents(ctx context.Context, events []types.Log, pars
 
 	store := ef.systemStore.WithTx(tx)
 
-	txnsEventsExistInDB := map[common.Hash]bool{}
+	persistedTxnHashEvents := map[common.Hash]bool{}
```
jsign, line 378: Just improving the name.

pkg/eventprocessor/eventfeed/impl/eventfeed.go
```diff
@@ -426,7 +427,11 @@ func (ef *EventFeed) persistEvents(ctx context.Context, events []types.Log, pars
 			ChainID:   ef.chainID,
 			EventJSON: eventJSONBytes,
 			EventType: eventType,
-		})
+		}
+		tblEvents = append(tblEvents, tblEvent)
+		if err := telemetry.Collect(ctx, toNewTablelandEvent(tblEvent)); err != nil {
+			return fmt.Errorf("collecting new tableland event metric: %s", err)
+		}
```
jsign, line 434: Same logic as mentioned before. We first save as a metric, and then as an event.

pkg/telemetry/storage/db_test.go
```diff
@@ -45,11 +45,7 @@ func TestCollectAndFetchAndPublish(t *testing.T) {
 			require.Equal(t, telemetry.StateHashType, metric.Type)
 			require.Equal(t, 1, metric.Version)
 			require.False(t, metric.Timestamp.IsZero())
-
-			sh := metric.Payload.(*telemetry.StateHashMetric)
-			require.Equal(t, fakeStateHash.ChainID, sh.ChainID)
-			require.Equal(t, fakeStateHash.BlockNumber, sh.BlockNumber)
-			require.Equal(t, fakeStateHash.Hash, sh.Hash)
+			require.Equal(t, &fakeStateHash, metric.Payload.(*telemetry.StateHashMetric))
```
jsign, line 48: I simplified some `.Equal` assertions in existing tests.

pkg/telemetry/storage/db_test.go
```diff
@@ -178,12 +188,92 @@ func TestCollectAndFetchAndPublish(t *testing.T) {
...
+		require.NoError(t, telemetry.Collect(context.Background(), metric))
+
+		metrics, err := s.FetchMetrics(context.Background(), false, 10)
+		require.NoError(t, err)
+		require.Len(t, metrics, 1)
+
+		require.Equal(t, telemetry.NewTablelandEventType, metrics[0].Type)
+		require.Equal(t, 1, metrics[0].Version)
+		require.False(t, metrics[0].Timestamp.IsZero())
+		require.Equal(t, &metric, metrics[0].Payload.(*telemetry.NewTablelandEventMetric))
```
jsign, line 276: New tests for both cases.

pkg/telemetry/storage/db_test.go
```diff
@@ -247,31 +337,9 @@ func TestCollectAndFetchAndPublish(t *testing.T) {
...
-	Version:      telemetry.ReadQueryMetricV1,
-	IPAddress:    "0.0.0.0",
-	SQLStatement: "SELECT * FROM foo",
-	FormatOptions: telemetry.ReadQueryFormatOptions{
-		Extract: true,
-		Unwrap:  false,
-		Output:  "objects",
-	},
-	TookMilli: 100,
-}
```
jsign, line 277: Moved to be define inside sub-tests if possible to avoid globals.

pkg/telemetry/telemetry.go
```diff
@@ -49,70 +49,30 @@ func Collect(ctx context.Context, metric interface{}) error {
 		return nil
 	}
 
+	var metricType MetricType
 	switch v := metric.(type) {
 	case StateHashMetric:
```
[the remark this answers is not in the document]
brunocalza: nice 👍

brunocalza (approved):
Awesome! Any ideas on how we could join the historical data and the new data coming from metrics? One approach I've applied in the past is to upload the historical data from the backup to BigQuery in a new table. Then, create a view in BigQUery that joins the historical data and the system metrics table.

jsign:
> Awesome! Any ideas on how we could join the historical data and the new data coming from metrics? One approach I've applied in the past is to upload the historical data from the backup to BigQuery in a new table. Then, create a view in BigQUery that joins the historical data and the system metrics table.

Yes, we can do that, but I think it might not be needed since soon we'll have to do a cold sync after fixing the two non-deterministic spec problems. In cold sync, all the history would be reconstructed again.

PR #396, opened 2022-11-10
First Ship meta PR

jsign:
# Summary

This PR has become the base branch for the next Ship.
It includes:
- Testnet/Mainnet network separations
- New SQL Spec features and fixes.
- The new API V1.
- Other general fixes/improvements.

# Context

See #287. 

# Implementation overview

NA since this has become a meta-PR.

# Implementation details and review orientation

NA

PR #402, opened 2022-11-18
Use mainnet url in client

asutula:


pkg/client/chains.go
```diff
@@ -5,6 +5,7 @@ import (
 )
 
 const (
+	mainnetURL = "https://tableland.network"
 	testnetURL = "https://testnet.tableland.network"
```
asutula, line 9: Just want to be sure... we need to update the testnet url at some point in the future, correct? Or did we find a way to let the existing testnet url remain the same?

jsign: See [here](https://www.notion.so/textile/Tableland-Testnet-Mainnet-separation-68c67911e68b410291cc03c911b97655#e647b9d032ed4c2b8d21f1976d03ebfe).

jsign:
> _ No description provided. _

Do you mind targeting this PR to [this branch](https://github.com/tablelandnetwork/go-tableland/pull/396)?
So whenever we merge that PR and we have the official split, this goes bundled with it.

pkg/client/chains.go
jsign: Notice also what's described in that document regarding the testnet network.

asutula: Ok so my interpretation of that is I should update this PR to use `testnets.tableland.network` for the `testnetURL`, correct?

jsign: Yep

asutula:
Ok @jsign if you want to give me a final approval, I'll merge this.

jsign (approved):

PR #411, opened 2022-12-07
bumps parser version

brunocalza:
# Summary

Bumps parser version to `de3c827`

# Context

It includes features such as all [`JOIN` operators](https://github.com/tablelandnetwork/go-sqlparser/pull/33), [compound queries](https://github.com/tablelandnetwork/go-sqlparser/pull/34), and [`INSERT` with `SELECT`](https://github.com/tablelandnetwork/go-sqlparser/pull/36).

# Implementation overview

- There was a refactoring in how the query validation is done to use the helpers from the parser package. The reason for that is so that the validator and the SDK can both rely on the same validation
- It was added the extra validation for an `INSERT` with `SELECT` query that checks if the chain of the insert table is the same chain of the select table, finishing the feature

### DIFF
--- a/pkg/eventprocessor/impl/eventprocessor_replayhistory_test.go
+++ b/pkg/eventprocessor/impl/eventprocessor_replayhistory_test.go
@@ -36,12 +36,12 @@ func TestReplayProductionHistory(t *testing.T) {
 
 	expectedStateHashes := map[tableland.ChainID]string{
 		1:      "ce4f083e256d3458a329b6cd1ba7d8e93d9703b3",
-		5:      "9ab5a92ac961d39ee07dd1e59518c527b39719c1",
-		10:     "3d907df6936eea9ab046d89c80b995d5db26e6ac",
+		5:      "b42740b92e40d616061a88872e19616c05469750",
+		10:     "c9afca82a8c30fc683da366b5f4920e7c131ec5d",
 		69:     "643af9ad784444242c6ef415727203941a720197",
 		137:    "b5fb42f3538738ab5856abf9e3b2e38d82378ca4",
-		420:    "8673f7095da19c1fc51f1f84b4c4ffc115a176e0",
-		80001:  "e85654e780c5e98cef2a25cd8e3764f125b2e1d1",
+		420:    "184800f533f4edd186853a85829fad8bc7802c4e",
+		80001:  "df269d9451e0bf1b77c7616aff4148a3f1a4559a",
 		421613: "073727932afcee9a5dba19f43c023689ca855dc2",
 	}
 


pkg/parsing/impl/validator.go
```diff
@@ -62,33 +61,24 @@ func (pp *QueryValidator) ValidateCreateTable(query string, chainID tableland.Ch
 	}
 
 	node := stmt.(*sqlparser.CreateTable)
-	if !pp.createTableNameRegEx.MatchString(node.Table.String()) {
-		return nil, &parsing.ErrInvalidTableName{}
-	}
-	parts := strings.Split(node.Table.String(), "_")
-	if len(parts) < 2 {
-		return nil, fmt.Errorf("table name isn't referencing the chain id")
+	validTable, err := sqlparser.ValidateCreateTargetTable(node.Table)
```
brunocalza, line 64: Using the parser package method to validate the table of a CREATE statement

pkg/parsing/impl/validator.go
```diff
@@ -398,24 +353,52 @@ func (s *readStmt) GetQuery() (string, error) {
...
+		return nil, fmt.Errorf("no system-table reference: %w", err)
+	}
+
+	insertTable, err := sqlparser.ValidateTargetTable(stmt.GetTable())
+	if err != nil {
+		return nil, fmt.Errorf("table name is not valid: %w", err)
 	}
 
-	return stmt.GetTable().String(), nil
+	if insert, ok := stmt.(*sqlparser.Insert); ok && insert.Select != nil {
```
brunocalza, line 366: If it is an INSERT with SELECT do the chain check

pkg/parsing/impl/validator_test.go
```diff
@@ -265,12 +265,14 @@ func TestWriteQuery(t *testing.T) {
 			expErrType: ptr2ErrInvalidSyntax(),
 		},
 
-		// Disallow JOINs and sub-queries
+		// insert with select chain mismatch
 		{
-			name:       "insert subquery",
```
brunocalza, line 270: This is the test to check for the chain mismatch error

pkg/parsing/impl/validator_test.go
```diff
@@ -166,7 +166,7 @@ func TestWriteQuery(t *testing.T) {
 		{
 			name:       "table id or chain id is missing",
 			query:      "delete from Hello_4 where a=2",
-			expErrType: ptr2ErrInvalidTableName(),
+			expErrType: ptr2ErrWrongFormatTableName(),
```
brunocalza, line 169: gotta change the error to use the error returned by the parser package

pkg/parsing/impl/validator.go
brunocalza: Most of the changes in this file is do adjust the code to the fact that the validation methods return the types `ValidatedTable` and `ValidatedCreateTable`

pkg/eventprocessor/impl/eventprocessor_replayhistory_test.go
```diff
@@ -36,12 +36,12 @@ func TestReplayProductionHistory(t *testing.T) {
 
 	expectedStateHashes := map[tableland.ChainID]string{
 		1:      "ce4f083e256d3458a329b6cd1ba7d8e93d9703b3",
-		5:      "9ab5a92ac961d39ee07dd1e59518c527b39719c1",
-		10:     "3d907df6936eea9ab046d89c80b995d5db26e6ac",
+		5:      "b42740b92e40d616061a88872e19616c05469750",
```
brunocalza, line 39: i'm not really sure why a change in the state happened. maybe an `INSERT` with `SELECT` query was accepted?

jsign (approved):
LGTM.

Left some comments reg the e2e test. I suggest some idea to try to get more confidence if there's something hairy going on. If we aren't sure what's happening, that's a bit concerning since we might discover in production.

PR #412, opened 2022-12-07
APIV1 implementation: Server and Go SDK

### LINKED ISSUE #358: The validator implements the new versioned HTTP API

### LINKED ISSUE #359: The Go SDK implements the new versioned HTTP API

jsign:
# Summary

This PR implements the defined [API V1 spec](https://github.com/tablelandnetwork/docs/pull/39).
We still keep JSON-RPC support since we're planning to deprecate it after some weeks of API V1 being available.

Fixes #358 
Fixes #359 
cc https://github.com/tablelandnetwork/docs/issues/7

Pendings:
- [x] [Wait for the decision around returning 404 when a table isn't found](https://github.com/tablelandnetwork/docs/pull/39#issuecomment-1339936913), adjust as necessary.
- [x] Wait until the final spec is merged into the https://github.com/tablelandnetwork/docs repository, and change the `Makefile` to pull from there.

# Context

As part of our goal to move out from JSON-RPC, we decided to design a new versioned REST API interface.

# Implementation overview

There're many things to comment, so I'll separate in sections.

## Implementation strategy regarding JSON-RPC support.
As mentioned in the summary section, in this PR we still need to keep JSON-RPC support for a while, which will be deprecated after some weeks of API V1 being live. 

With that in mind, the implementation of this PR is optimized to make dropping JSON-RPC shortly as simple as possible. I clarify this because if JSON-RPC support were going to be an official API supported for much longer, I'd probably prefer to structure some things differently. Consider that when reviewing how things would look after we drop support since that will be the end state of the codebase in the short term.

## CI/CD

The CI/CD for deployment uses the APIs to determine if the deployment was done correctly but does some liveness probes. These were changed to use the new APIs.

## OpenAPI code generation

In hindsight, this hasn't been a great experience, but I tried as much as possible to save the idea. The good part is that I found, what I consider, a reasonable tradeoff to allow the autogenerated code to be untouched since breaking that rule is a headache when regenerating.

The TL;DR is that doing `make gen-api-v1` will pull the spec from the docs repository, autogenerate the code, remove unneeded files, and make any further code change in the script. There's no manual work, and everything is explicit in the Makefile rule. 

More about this in the PR comments.

## Go clients

This PR contains a new Go client for APIV1, but also keeps the legacy Go client until we drop JSON-RPC support. These clients are organized in `client/v1` and `client/legacy`, respectively. When dropping support for JSON-RPC, we can delete the latter.

## API e2e testing

The validator has a lot of internal e2e testing and great coverage. That wasn't the case for the JSON-RPC HTTP layer. We had tests, but they relied on mocks which is fine, but they aren't e2e tests. This was a good opportunity to change that.

I took advantage of the new Go client implementation and created a test that covers all the behavior defined in the spec. That is all the mentioned endpoints and their respective potential status code cases (e.g., 200, 404, 400).

The existing tests for APIs are still there, we can decide at some point if they add value or add friction to the development process if we feel we're covered with these new e2e tests. Probably when dropping support for JSON-RPC we can delete some.

## Pulling stalled https://github.com/tablelandnetwork/go-tableland/pull/279

I pulled the test harness idea from https://github.com/tablelandnetwork/go-tableland/pull/279 that got stalled (PR closed). I had to make multiple adjustments since multiple things changed after that PR, but the essence and a big chunk of the code is the same.

cc @asutula 

## API code versioning

We currently support APIv1 and the legacy JSON-RPC that will be dropped soon. 
The way this was organized was:
- `internal/router/controllers/apiv1`: contains everything related to the API V1. In this case, the autogenerated code (models, and boilerplate)
- `internal/router/controllers/legacy`: contains everything related to the legacy API (i.e: JSON-RPC)

When having further v2, v3, etc, we can keep this folder structure. I have the sensation that whenever we have at least two long-term version to be supported (probably v1 and v2), we might want to tune some things further.

I'd recommend reviewing in the same order as listed in the "Files changed" tab. I think that's a reasonable path to walk. Only in a particular moment close to the end I'll recommend jumping somewhere else and then coming back. 

# Checklist


tests/fullstack/fullstack.go
```diff
@@ -0,0 +1,207 @@
+package fullstack
```
jsign, line 1: Inspired by the original PR, but I had to fix a bunch of things reg things that happened since then; but the essence is the same.

pkg/sqlstore/impl/system/store.go
```diff
@@ -246,7 +246,7 @@ func (s *SystemStore) GetSchemaByTableName(ctx context.Context, name string) (sq
 		return sqlstore.TableSchema{}, fmt.Errorf("failed to get the table: %s", err)
 	}
 
-	index := strings.LastIndex(createStmt, "strict")
+	index := strings.LastIndex(strings.ToLower(createStmt), "strict")
```
jsign, line 249: @brunocalza, when deploying to testnet I noticed that this caused a panic.
After some digging, it looks like for the older tables it returned `STRICT` (uppercase).

Probably this is related that in our previous parser we had some upper cases than we changed in the latest versions. And, probably SQLite stores in `sqlite_schema` the original statement executed. 

So this means that we'll have both cases in `sqlite_schema`:
1- All the existing tables before this parser upgrade that used uppercase
2- The new tables will use lowercase.

This is my theory of the explanation to this panic.

As you see, the fix is quite easy. But it's interesting to note that we should remember this about everything we do in `sqlite_schema` from now on, since this "uppercase/lowercase" situation can backfire again in the future. Probably the best way to deal with this is always remember to lowercase (or uppercase) before trying to do any "unstructured finding" (like `LastIndex` or similar).

pkg/client/v1/client.go
```diff
@@ -0,0 +1,193 @@
...
+		systemimpl.RegistryTableName,
+	}, parserOpts...)
+	if err != nil {
+		return nil, fmt.Errorf("new parser: %s", err)
+	}
+
+	parser, err = parserimpl.NewInstrumentedSQLValidator(parser)
+	if err != nil {
+		return nil, fmt.Errorf("instrumenting parser: %s", err)
+	}
```
asutula, line 122: Does it make sense to use a instrumented parser here? What does instrumentation do if the client is being used in a client side program (not within the validator)?

jsign: Yeah, might be too much to think about collecting metrics for a client.
I'll remove the instrumentation :+1:

internal/router/router.go
```diff
@@ -1,146 +1,187 @@
...
+		endpoint, ok := handlers[routeName]
+		if !ok {
+			return fmt.Errorf("route with name %s not found in handler", routeName)
+		}
+		pathTemplate, err := route.GetPathTemplate()
+		if err != nil {
+			return fmt.Errorf("get path template: %s", err)
+		}
+
+		router.get(
```
brunocalza, line 141: You are calling `router.get` here because all endpoints defined in [`routes`](https://github.com/tablelandnetwork/go-tableland/pull/412/files#diff-be4fcea6bb23790b7d2da94d48596eec4c96e6c507124ad110278d08b658fc50R50) respond to the `GET` method, right? 

Looks like to me that if we generate a `POST` route, this would code would not work correctly or I'm missing something

internal/router/router.go
```diff
@@ -1,146 +1,187 @@
...
+			endpoint.handler,
+			append(endpoint.middlewares, middlewares.OtelHTTP(routeName))...,
+		)
+		return nil
+	}); err != nil {
+		return fmt.Errorf("configuring api v1 router: %s", err)
+	}
+	if specRoutesCount != len(handlers) {
+		return fmt.Errorf("the spec has less defined routes than expected handlers to be used")
+	}
```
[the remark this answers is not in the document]
brunocalza: Really interesting approach. Good work!

brunocalza (approved):
Awesome work! Looks pretty good to me! I left a couple of comments worth checking

Makefile
```diff
@@ -79,7 +76,21 @@ test-replayhistory:
...
+.PHONY: lint
+
+# OpenAPI
+SPEC_URL=https://raw.githubusercontent.com/tablelandnetwork/docs/jsign/vh/specs/validator/tableland-openapi-spec.yaml
+APIV1=${PWD}/internal/router/controllers/apiv1
+gen-api-v1:
+	mkdir -p ${APIV1}
+	curl -s ${SPEC_URL} > ${APIV1}/tableland-openapi-spec.yaml
+	docker run -w /gen -e GEN_DIR=/gen -v ${APIV1}:/gen swaggerapi/swagger-codegen-cli-v3 \
+	   generate --lang go-server -o /gen -i tableland-openapi-spec.yaml --additional-properties=packageName=apiv1
```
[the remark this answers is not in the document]
jsign: Sounds like a good idea, I'll tag the latest version.

internal/router/router.go
jsign: Correct, since all the methods are GET, this is fine.
Whenever another verb appears, we can use `.GetMethods()` and extend this.

But technically, that should happen in a v2 that would be new code. v1 in theory can't have those kind of changes so this `.get(..)` should be safe to be permanent.

PR #417, opened 2022-12-13
controller/v1: return content-type when body isn't empty

jsign:
# Summary

Return `Content-Type: application/json` in responses that have a JSON body.

# Context

# Implementation overview

NA

NA


carsonfarmer (approved):
Awesome, yup this'll do just nicely!

PR #418, opened 2022-12-14
CI: generate binaries and Docker images

### LINKED ISSUE #365: Implement CI/CD actions to create releases to the repository when a tag is created

jsign:
# Summary

This PR includes a new Github Action that is run when a tag is created, which:
- Creates binaries for `linux/amd64`, `windows/amd64`, and `darwin/amd64`
- Create Docker images for `amd64` and `arm64`.

Closes #365.

**Note: This PR is ready for review, but I'll keep it in draft until Bruno returns.**

# Context

This is part of our work to provide more options to validators apart from building from source.

# Implementation overview

Building these binaries isn't as easy as using the cross-compiling feature of Go, since we rely on CGO, which complicates things quite a bit. In particular, for M1 binaries since Github Actions doesn't have `arm64` VMs (at least, yet). I gave it a shot to a QEMU setup to simulate an `arm64` arch on top of `amd64` It was pretty good, but unfortunately, the GCC version that M1 needs is a custom XCode one, so it was getting too complicated. Probably with more time I can find some other workaround if there's some Docker image with this GCC compiler.. so I can use QEMU to simulate the arch.

We can wait for Github Actions to provide a M1 machine setup, or also use something that @asutula  [mentioned](https://discord.com/channels/592843512312102924/1013931507309617232/1052332699576189069) but that has some extra cost. Also, if users with M1 want binaries to run locally, I think somebody can run the `amd64` binary with Rosetta, and it should work too.

For the Docker images, they can be used in M1 since the build process uses QEMU to at least simulate a `arm64` architecture, which should mean that it _should_ be runnable in an M1 (I don't have an M1 machine to verify this yet). So, an M1 machine can run a `linux/arm64` (that's the theory). Building this `arm64` with QEMU takes 10x more time than the `amd64`, but I think that's fine for our CI.

TL;DR:
- If you want to see an example of binaries generated, [see this test release](https://discord.com/channels/592843512312102924/1013931507309617232/1052319113906765874).
- For our Docker images, [see this Dockerhub repo](https://hub.docker.com/repository/docker/textile/tableland/tags?page=1&ordering=last_updated). Notice how there're two architecture versions for the images to provide support for both.

NA

# Checklist

### DIFF
--- a/.github/workflows/binaries.yml
+++ b/.github/workflows/binaries.yml
@@ -1,0 +1,61 @@
+name: Generate binaries
+on:
+  release:
+    types:
+      - created
+jobs:
+  binaries:
+    strategy:
+      matrix:
+        include:
+          - os: ubuntu-latest
+            asset_name: api-linux-amd64.tar.gz
+            compress_cmd: tar -czvf
+            build_cmd: docker run -v $PWD:/data golang:1.19 bash -c "cd /data && go build ./cmd/api"
+          - os: windows-latest
+            asset_name: api-windows-amd64.zip
+            compress_cmd: tar.exe -a -c -f
+            build_cmd: go build -o api ./cmd/api
+          - os: macos-latest
+            asset_name: api-darwin-amd64.tar.gz
+            compress_cmd: tar -czvf
+            build_cmd: go build -o api ./cmd/api
+    runs-on: ${{ matrix.os }}
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@v3
+      - name: Install Go
+        uses: actions/setup-go@v2
+        with:
+          go-version: v1.19.x
+      - name: Build binary
+        run: ${{ matrix.build_cmd }}
+      - name: Pack output
+        run: ${{ matrix.compress_cmd }} ${{ matrix.asset_name }} api
+      - name: Upload binary
+        uses: svenstaro/upload-release-action@v2
+        with:
+          repo_token: ${{ secrets.GITHUB_TOKEN }}
+          file: ${{ matrix.asset_name }}
+          asset_name: ${{ matrix.asset_name }}
+          tag: ${{ github.ref_name }}
+          overwrite: true
+  docker:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v2
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v2
+      - name: Login to Docker Hub
+        uses: docker/login-action@v2
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+      - name: Build and push
+        uses: docker/build-push-action@v3
+        with:
+          file: ./cmd/api/Dockerfile
+          push: true
+          tags: textile/tableland:latest,textile/tableland:${{ github.ref_name }}
+          platforms: linux/amd64, linux/arm64
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -14,7 +14,7 @@ on:
 
 jobs:
   deploy:
-    if: github.event_name == 'release' || github.ref == 'refs/heads/main' || contains(github.event.head_commit.message, '[stagingdeploy]') || contains(github.event.head_commit.message, '[testnetdeploy]') || contains(github.event.head_commit.message, '[mainnetdeploy]')
+    if: contains(github.event.head_commit.message, '[stagingdeploy]') || contains(github.event.head_commit.message, '[testnetdeploy]') || contains(github.event.head_commit.message, '[mainnetdeploy]')
     name: Deploy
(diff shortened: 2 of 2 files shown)


.github/workflows/binaries.yml
```diff
@@ -0,0 +1,61 @@
+name: Generate binaries
+on:
+  release:
+    types:
+      - created
```
jsign, line 5: Only run in releases.

.github/workflows/binaries.yml
```diff
@@ -0,0 +1,61 @@
...
+jobs:
+  binaries:
+    strategy:
+      matrix:
+        include:
+          - os: ubuntu-latest
+            asset_name: api-linux-amd64.tar.gz
+            compress_cmd: tar -czvf
+            build_cmd: docker run -v $PWD:/data golang:1.19 bash -c "cd /data && go build ./cmd/api"
```
jsign, line 14: You might ask, why not: `go build ..` style?
That was the first approach, but unfortunately, the GLIBC version that Github Actions VMs are using is quite new, and I couldn't even run that binary on my Ubuntu desktop machine. So what I did here compile the binary inside the `golang:1.19` docker image that has a lower version of GLIBC that will be compatible with more machines.

This might be a good idea to be applied in the other binaries, but unless somebody reports this happening in other places  maybe we can avoid it.

.github/workflows/binaries.yml
```diff
@@ -0,0 +1,61 @@
+name: Generate binaries
+on:
+  release:
+    types:
+      - created
+jobs:
+  binaries:
```
jsign, line 7: This might see a bit "manual", but I've tried some other Github Actions that exist out there and they're mostly created to do cross-compiling in Go without needing CGO. So, I got actions that refused to run in Windows machines and things like that.

In any case, those actions aren't doing anything magical... I ended up doing all this manually which allow us to be more flexible anyway.

.github/workflows/binaries.yml
```diff
@@ -0,0 +1,61 @@
...
+          repo_token: ${{ secrets.GITHUB_TOKEN }}
+          file: ${{ matrix.asset_name }}
+          asset_name: ${{ matrix.asset_name }}
+          tag: ${{ github.ref_name }}
+          overwrite: true
+  docker:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v2
```
jsign, line 47: Just in case: QEMU is a machine emulator/virtualizer. So you can run binaries from other architectures in your architecture. Of course, the run will be way slower than running natively, but it's useful to do multi-platform image creations.

.github/workflows/binaries.yml
```diff
@@ -0,0 +1,61 @@
...
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+      - name: Build and push
+        uses: docker/build-push-action@v3
+        with:
+          file: ./cmd/api/Dockerfile
+          push: true
+          tags: textile/tableland:latest,textile/tableland:${{ github.ref_name }}
+          platforms: linux/amd64, linux/arm64
```
jsign, line 61: This line is a bit magical because `buildx` will leverage that QEMU is installed in the VM, and to build for architectures that aren't the native one in the VM, it will run all the Docker process using QEMU.

This means that the `Dockerfile` image building for `linux/arm64` will be run in QEMU (since we only have `amd64` machines in GHA). It takes 10x the time compared to `linux/amd64` which is run natively, but still 10mins is a reasonable time to wait.

Note that I tried the same idea for binaries, so I can run a `golang:1.19` in a simulated `arm64` machine. The problem was that I needed the GCC XCode compiler installed in Ubuntu to do the CGO cross compiling, which was nasty.
In Docker, M1 machines could run images that match for their architecture... that's why these docker images should be runnable in M1.

.github/workflows/deploy.yml
```diff
@@ -14,7 +14,7 @@ on:
 
 jobs:
   deploy:
-    if: github.event_name == 'release' || github.ref == 'refs/heads/main' || contains(github.event.head_commit.message, '[stagingdeploy]') || contains(github.event.head_commit.message, '[testnetdeploy]') || contains(github.event.head_commit.message, '[mainnetdeploy]')
+    if: contains(github.event.head_commit.message, '[stagingdeploy]') || contains(github.event.head_commit.message, '[testnetdeploy]') || contains(github.event.head_commit.message, '[mainnetdeploy]')
```
jsign, line 17: I disabled this until we confirm our release strategy [in our doc](https://www.notion.so/textile/Validator-Releases-Playbook-84f90706d20f487aa3dbbbda3d59d563).

Probably what we'll confirm is that we should only delete `github.ref == 'refs/heads/main' `.

.github/workflows/binaries.yml
brunocalza: Sounds good!

brunocalza (approved):
Looks good to me!

Issue #400, opened 2022-11-17
healthbot: stop relying on validators transaction relaying


jsign:
Today, Healthbot relies on the validator to relay transactions. This feature will become deprecated soon.

We need to change how heathbot works so it sends the transaction with its wallet using an Eth node provider.

PR #423, opened 2022-12-19
healthbot: stop using on validator transaction relaying

jsign:
# Summary

Closes https://github.com/tablelandnetwork/go-tableland/issues/400

# Context

In the upcoming API v1, validators won't be offering transaction relaying. Our current Healthbot implementation 

**Note: this PR is ready for review, but keeping it as draft until Bruno comes back**

# Implementation overview

The implementation goes along the lines of a previously stalled attempt #279 with some slight changes.

The infrastructure work we have to in each validator is to update the `deployed/<env>/healthbot/.env_healthbot` files removing the previous `HEALTHBOT_<chainName>_SIWE` and adding `HEALTHBOT_<chainName>_PRIVATE_KEY`.

NA

# Checklist


pkg/client/chains.go
```diff
@@ -40,64 +40,64 @@ var ChainIDs = struct {
 type Chain struct {
```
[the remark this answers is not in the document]
asutula: Ok sounds good. There's something I like about the primary entry point to referencing chains being a human readable name (`EthereumGoerli`, for example), but I'm fine with trying this out.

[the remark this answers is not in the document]
asutula: I realize mapping from strings in a config file to a struct member name is harder, more error prone, and more tedious. Using chain ids is definitely easier in that way.

asutula (approved):
Good work! Looks familiar, I like it ;)

docker/deployed/testnet/healthbot/config.json
```diff
@@ -6,41 +6,41 @@
         "Human": false,
         "Debug": true
     },
-    "Target": "https://testnet.tableland.network/rpc",
+    "Target": "https://testnets.tableland.network/rpc",
```
brunocalza, line 9: Should we remove the`Target` entry in this config? (also in `staging`)

brunocalza (approved):
Nice! The code became much simpler with the client :)

docker/deployed/testnet/healthbot/config.json
jsign: Good point, will do.

jsign: @brunocalza, done, added an extra commit. Ready to have another approval for merging.

brunocalza (approved):

PR #424, opened 2022-12-22
Support custom functions in read and write query

### LINKED ISSUE #352: Support EVM specific functions for read-queries and write-queries
There's been a few cases come up where it would be nice to grab some available state from transactions:

- `TXN_HASH()`: The validator would replace this text with the has of the txn that delivered the event (only available to write queries).
- `BLOCK_NUM()`: The validator would replace this text with the number of the block that delivered the event (only available to write queries).
- `BLOCK_NUM(<chain_id>)`: The validator would replace this text with the number of the _last seen_ block (only available to read queries). For example, in Rigs we want to show how long a rig has been _in-flight_ at a metadata attribute. We have `start_time` and `end_time` columns on a flight sessions table that is contains block numbers. Each flight session results in a row in this table with a start time. If `end_time` is `null`, the rig is in-flight. So, to calculate total flight time including an active session, we'd use this method:

```
SELECT SUM(COALESCE(end_time, BLOCK_NUM(<chain_id>))) - SUM(start_time) 
FROM pilot_sessions_80001_3515 WHERE rig_id=11
```

`BLOCK_TIME` could be added in the same way as `BLOCK_NUM`, with a read and write flavor.

jsign:
# Summary

Closes #352.

**Note: This PR is still in draft mode because it depends on https://github.com/tablelandnetwork/go-sqlparser/issues/14. It's using functions that don't exist yet (so not compiling is expected). After we have the parser work, we can wrap things here.**

# Context

This work is part of a protocol upgrade that will allow custom functions in read and write queries.

In summary, the plan is to:
- Write queries can use `TXN_HASH()` and `BLOCK_NUM()`.
- Read queries can use `BLOCK_NUM(<chain-id>)`
For more details, see #352.

# Implementation overview

This PR is to include in the execution pipeline a step to instantiate potentially custom function calls in read or read queries.

The high-level idea is that parsed queries `GetQuery()` method is now upgraded to `GetQuery(resolver ReadQueryResolver) string` and `GetQuery(resolver WriteQueryResolver) string` for read and write queries, respectively.

The `<Read|Write>QueryResolver` interface provides the parser with functions to resolve any supported function. The idea is that the parser will walk the AST, and whenever it finds, for example, a `TXN_HASH()` call, it can call the `resolver.GetTxnHash()`, which will provide a `string` that can replace the function call node AST.

This means that an AST with nodes calling the supported custom functions would be transformed into another where these calls were replaced with their corresponding literals. 

The implementor of `<Read|Write>QueryResolver` is the validator execution pipeline since it's the party that knows where the query is being executed (i.e: the "context" of execution).

If we want to add more custom functions in the future, we add new methods to `<Read|Write>QueryResolver`, and the mechanics would be the same.

Note that read and write queries have different resolver interfaces (i.e: `ReadQueryResolver`, `WriteQueryResolver). This is necessary to be clear that both read and write queries support different functions.

I'll be adding PR comments even if this will be in Draft until some other work happens, it will help understand how things are happening.

I'd recommend the following order:
1. `pkg/parsing/query_validator.go`: to understand some internal changes in prepared statements and resolver interfaces. This is necessary to understand changes in the user store wiring and the execution pipeline.
2. `pkg/parsing/impl/validator.go`: how the resolver is wired into the parsed statements. This might be the most important part to inspect to think about the `go-sqlparser` related changes.
3. `pkg/readqueryresolver/readqueryresolver.go`: to see how the `ReadQueryResolver` interface implementation, which is quite simple.
4. `pkg/sqlstore/impl/user/store.go`: to see how the resolver now is wired in `UserStore` which is the "entry point" for executing read queries.
5. `pkg/eventprocessor/impl/executor/impl/blockscope.go`: to see the `WriteQueryResolver` implementation in the execution pipeline.
6. `pkg/eventprocessor/impl/executor/impl/txnscope_runsql.go`: to see where the "magic"/gluing happens.

After that, you'll understand the gist of everything. You can review the rest and understand secondary changes or wirings from there.

I'll assume the review is being done in the suggested order, so I can assume the reviewer already knows some core things while commenting wirings/tests/etc.

# Checklist


dtbuchholz:
@jsign is there a reason why reads would need to use `BLOCK_NUM(<chain-id>)` instead of `BLOCK_NUM()`? E.g., could the `chain_id` be inferred based on the table it’s reading from?

jsign:
> @jsign is there a reason why reads would need to use `BLOCK_NUM(<chain-id>)` instead of `BLOCK_NUM()`? E.g., could the `chain_id` be inferred based on the table it’s reading from?

Not always, since a read-query can be joining tables from different chains, or even no tables at all (e.g: `select block_num()` what would mean?). In some read query cases that only join tables from a single chain, then we could infer it. But in many other cases `BLOCK_NUM()` won't work. Note that `BLOCK_NUM()` exists for write-queries, since in that context you know that they always are scoped to a particular chain.

As a suggestion, it's better to have these discussions at https://github.com/tablelandnetwork/go-tableland/issues/352 which is the issue that originally described the need.

pkg/parsing/impl/validator_test.go
```diff
@@ -116,7 +116,7 @@ func TestReadRunSQL(t *testing.T) {
 				if tc.expErrType == nil {
 					require.NoError(t, err)
 					require.NotNil(t, rs)
-					q, err := rs.GetQuery()
+					q, err := rs.GetQuery(nil)
```
jsign, line 119: We can decide if we want to allow passing a `nil` resolver or not.
The reality, is that in the real validator implementation a non-nil value is always passed. Maybe for tests is useful.

So we have two options:
1. We allow `nil` and if `go-sqlparser` detects that the query has some `TXN_HASH()` but the resolver is `nil`, it can return an error like "resolver needed" or similar. But if that query doesn't have any custom function, it isn't a problem to have received a `nil` resolver because it wasn't needed.
2. We don't allow `nil` values, and we fail early at the `go-validator` level. For tests, we can create a dummy resolver that does nothing or panics.

I don't have a particularly strong opinion, since this is mostly for tests.

pkg/parsing/impl/validator_test.go
```diff
@@ -891,6 +894,163 @@ func TestWriteStatementAddReturningClause(t *testing.T) {
...
+					require.Error(t, err)
+					return
+				}
+				require.Equal(t, tc.expQuery, q)
+			}
+		}(it))
+	}
+}
+
+func TestCustomFunctionResolveWriteQuery(t *testing.T) {
```
jsign, line 945: Covering different angles of write-queries using many custom functions at once.

tests/fullstack/fullstack.go
```diff
@@ -19,13 +19,15 @@ import (
 	systemimpl "github.com/textileio/go-tableland/internal/system/impl"
```
jsign, line 19: Nothing much interesting here. I had to move around some wires since the user store now depends on the event processors for creating resolvers. Mostly boilerplate stuff.

pkg/parsing/query_validator.go
```diff
@@ -9,16 +9,25 @@ import (
...
-type Stmt interface {
-	GetQuery() (string, error)
+// WriteQueryResolver resolves Tablealand Custom Functions for a write statement.
+type WriteQueryResolver interface {
+	// GetTxnHash returns the transaction hash of the transaction containing the query being processed.
+	GetTxnHash() string
+
+	// GetBlockNumber returns the block number of the block containing query being processed.
+	GetBlockNumber() int64
+}
```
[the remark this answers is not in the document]
brunocalza: I think the `WriteQueryResolver` and `ReadQueryResolver` should be defined in the `go-sqlparser` to avoid `go-sqlparser` depending on `go-tabeland`. 

That would change [L44](https://github.com/tablelandnetwork/go-tableland/pull/424/files#diff-c6920746c771dcedad8850e6ba75a3e2b487df3092caab3b6a8ee9dbb0cf5f50R44) to:
`GetQuery(sqlparser.WriteQueryResolver) (string, error)`

pkg/parsing/impl/validator.go
```diff
@@ -220,8 +220,12 @@ type mutatingStmt struct {
 
 var _ parsing.MutatingStmt = (*mutatingStmt)(nil)
 
-func (s *mutatingStmt) GetQuery() (string, error) {
-	return s.node.String(), nil
+func (s *mutatingStmt) GetQuery(resolver parsing.WriteQueryResolver) (string, error) {
+	query, err := s.node.Resolve(resolver)
```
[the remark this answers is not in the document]
brunocalza: >The go-sqlparser will define its own WriteQueryResolver interface, and we should use that one here instead of the one defined in go-validator.

Oh, forget about [this](https://github.com/tablelandnetwork/go-tableland/pull/424/files#r1057327426) comment.

pkg/parsing/impl/validator_test.go
brunocalza: 1 sounds good!

go.mod
```diff
@@ -22,7 +22,7 @@ require (
 	github.com/spf13/cobra v1.6.0
 	github.com/spruceid/siwe-go v0.2.1-0.20220804171946-fc1b0374f4ff
 	github.com/stretchr/testify v1.8.1
-	github.com/tablelandnetwork/sqlparser v0.0.0-20221209125022-863bbd9f6527
+	github.com/tablelandnetwork/sqlparser v0.0.0-20221229133649-a34f5eedb409
```
brunocalza, line 25: This is the PR's latest commit. If every thing is ok, I'll merge that PR, then update this to the main latest commit.

pkg/eventprocessor/impl/executor/impl/txnscope_runsql_customfuncs_test.go
```diff
@@ -78,13 +78,6 @@ func TestCustomFunctionsWriteQuery(t *testing.T) {
 			newExecutorWithTable: newExecutorWithIntegerTable,
 			assertExpectation:    checkBlockNumberEq1,
 		},
-		{
```
brunocalza: I deleted this test because it fails at different levels than other tests, and the `mustFail` does not give context to the assertion logic. This case is already tested in the parser, so I think it's fine.

pkg/eventprocessor/impl/executor/impl/txnscope_runsql_customfuncs_test.go
```diff
@@ -99,12 +92,13 @@ func TestCustomFunctionsWriteQuery(t *testing.T) {
...
 				ex, dbURI := test.newExecutorWithTable()
 
-				bs, err := ex.NewBlockScope(ctx, 0)
+				bs, err := ex.NewBlockScope(ctx, 1)
 				require.NoError(t, err)
 
 				txnHash, res, err := execTxnWithRunSQLEvents(t, bs, []string{test.query})
 				if test.mustFail {
-					require.Error(t, err)
+					require.NotNil(t, res.Error)
```
brunocalza, line 100: Changed this because the errors are actually syntax errors and come through the `res`.

pkg/parsing/impl/validator.go
```diff
@@ -220,12 +220,16 @@ type mutatingStmt struct {
 
 var _ parsing.MutatingStmt = (*mutatingStmt)(nil)
 
-func (s *mutatingStmt) GetQuery(resolver parsing.WriteQueryResolver) (string, error) {
-	query, err := s.node.Resolve(resolver)
-	if err != nil {
-		return "", fmt.Errorf("resolving write query: %s", err)
+func (s *mutatingStmt) GetQuery(resolver sqlparser.WriteStatementResolver) (string, error) {
+	if writeStmt, ok := s.node.(sqlparser.WriteStatement); ok {
```
brunocalza, line 224: Gotta make this check because a `MustatingStm` can be a `Grant` or `Revoke`

pkg/parsing/impl/validator_test.go
```diff
@@ -59,24 +59,8 @@ func TestReadRunSQL(t *testing.T) {
...
-					  json_object('trait_type', 'Background', 'value', rigs.background),
-					  json_object('trait_type', (
-						select name from badges_2 where badges.rig_id = rigs.id and position = 1 limit 1
-					  ), 'value', (
-						select image from badges_2 where badges.rig_id = rigs.id and position = 1 limit 1
-					  ))
-				  )
-			  )
-			  from rigs_1;`,
+			name:       "select with complex subqueries in function arguments",
```
brunocalza, line 62: For some reason, the test was failing because of the indentation. Really weird

pkg/parsing/impl/validator.go
```diff
@@ -353,10 +357,10 @@ type readStmt struct {
 
 var _ parsing.ReadStmt = (*readStmt)(nil)
 
-func (s *readStmt) GetQuery(resolver parsing.ReadQueryResolver) (string, error) {
-	query, err := s.statement.Resolve(resolver)
+func (s *readStmt) GetQuery(resolver sqlparser.ReadStatementResolver) (string, error) {
+	query, err := s.statement.(sqlparser.ReadStatement).Resolve(resolver)
```
brunocalza, line 361: this cast should not be a problem, but maybe worth a check

brunocalza (approved):

PR #427, opened 2022-12-30
Updates parser with new custom function implementation

brunocalza:
# Summary

Updates parser with new custom function implementation and improve the error handling

# Context

- See https://github.com/tablelandnetwork/go-sqlparser/pull/43
- Also, when a query cannot be resolved we need to return `errorQueryExecution` error so that the execution pipeline does not halt


jsign (approved):
LGTM.

Maybe worth some `[testnetdeploy]` to update the new testnet validator?

PR #430, opened 2023-01-03
fix: get schema was not parsing statements with autoincrement 

### LINKED ISSUE #428: Error while fetching table metadata
The call `https://testnets.tableland.network/api/v1/tables/420/198` returns

```
{"message":"Failed to fetch metadata"}
```

The log message says:

```
get table schema information: failed to parse create stmt: syntax error at position 81 near 'autoincrement'
```

This is probably because the table has a column `INTEGER PRIMARY KEY`. That implied an additional `AUTOINCREMENT` to the table definition statement. When we ask for table metadata, we get the table definition statement and try to parse does not understand the `AUTOINCREMENT` keyword.

You can check the schema of the table with the call: `https://testnets.tableland.network/query?s=select%20sql%20from%20sqlite_master%20WHERE%20sql%20like%20%27%25420_198%25%27`

brunocalza:
# Summary

When getting the schema of a table we need to parse the statement that is stored in SQLite. Sometimes that statement may come with an `AUTOINCREMENT` (because of the protocol rules that add the `AUTOINCREMENT` keyword automatically when an `INTEGER PRIMARY KEY` is found). However, the `AUTOINCREMENT` keyword is not parsable. So we're getting an error. 

# Context

Closes #428 

# Implementation overview

Simply remove the `AUTOINCREMENT` before parsing.

I don't think removing keywords that are not parsable such as `STRICT` and `AUTOINCREMENT` is a good long-term approach. But this is good enough to solve the `bug`. 

This is also related to https://github.com/tablelandnetwork/go-sqlparser/issues/41. It is kind of weird that the `String` method produces and string that is not compliant with the Tableland SQL. But that's more a matter of definition. Up until now, we've been using the `String` method as a way to get the SQL statement that will be executed in `SQLite`. I'll leave this discussion for the future.

### DIFF
--- a/pkg/sqlstore/impl/system/store.go
+++ b/pkg/sqlstore/impl/system/store.go
@@ -246,6 +246,10 @@ func (s *SystemStore) GetSchemaByTableName(ctx context.Context, name string) (sq
 		return sqlstore.TableSchema{}, fmt.Errorf("failed to get the table: %s", err)
 	}
 
+	if strings.Contains(strings.ToLower(createStmt), "autoincrement") {
+		createStmt = strings.Replace(createStmt, "autoincrement", "", -1)
+	}
+
 	index := strings.LastIndex(strings.ToLower(createStmt), "strict")
 	ast, err := sqlparser.Parse(createStmt[:index])
 	if err != nil {
--- a/internal/system/impl/sqlstore_test.go
+++ b/internal/system/impl/sqlstore_test.go
@@ -128,7 +128,7 @@ func TestGetSchemaByTableName(t *testing.T) {
 			&ethereum.ContractCreateTable{
 				TableId:   big.NewInt(42),
 				Owner:     common.HexToAddress("0xb451cee4A42A652Fe77d373BAe66D42fd6B8D8FF"),
-				Statement: "create table foo_1337 (a int primary key, b text not null default 'foo' unique, check (a > 0))",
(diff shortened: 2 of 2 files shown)


pkg/sqlstore/impl/system/store.go
```diff
@@ -246,6 +246,10 @@ func (s *SystemStore) GetSchemaByTableName(ctx context.Context, name string) (sq
 		return sqlstore.TableSchema{}, fmt.Errorf("failed to get the table: %s", err)
 	}
 
+	if strings.Contains(strings.ToLower(createStmt), "autoincrement") {
+		createStmt = strings.Replace(createStmt, "autoincrement", "", -1)
+	}
```
jsign, line 251: Uhm, OK, looks fine, but I'm a bit worried we can replace any other stuff with `autoincrement` that shouldn't be replaced. For example, if a column name is `autoincrement`, something bad would happen. 

Of course, this is a very weird border case, but still possible.

I can suggest two options:
1- Simply accept that risk, since it would be very weird if this happened; and wait for a better overall solution, probably related to the other chat about generating "spec compatible" strings.
2- We can do a similar strategy as with `strict`, looking for a more concrete position to replace. For example, maybe look for `primary key autoincrement` (not sure if that would be correct, but you get the idea), and only replace it there. That will probably get rid of this problem while still being a simple-ish solution.

I'm fine with either.

jsign (approved):

PR #433, opened 2023-01-17
Github Action deployment flow fixes

jsign:
# Summary

This PR fixes the rules to decide when the action runs, and internal job would apply.

# Context

NA

# Implementation overview

NA

NA

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -14,7 +14,7 @@ on:
 
 jobs:
   deploy:
-    if: contains(github.event.head_commit.message, '[stagingdeploy]') || contains(github.event.head_commit.message, '[testnetdeploy]') || contains(github.event.head_commit.message, '[mainnetdeploy]')
+    if: github.event_name == 'release' || github.ref == 'refs/heads/main' || contains(github.event.head_commit.message, '[stagingdeploy]') || contains(github.event.head_commit.message, '[testnetdeploy]') || contains(github.event.head_commit.message, '[mainnetdeploy]')
     name: Deploy
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy.yml
```diff
@@ -14,7 +14,7 @@ on:
 
 jobs:
   deploy:
-    if: contains(github.event.head_commit.message, '[stagingdeploy]') || contains(github.event.head_commit.message, '[testnetdeploy]') || contains(github.event.head_commit.message, '[mainnetdeploy]')
+    if: github.event_name == 'release' || github.ref == 'refs/heads/main' || contains(github.event.head_commit.message, '[stagingdeploy]') || contains(github.event.head_commit.message, '[testnetdeploy]') || contains(github.event.head_commit.message, '[mainnetdeploy]')
```
jsign, line 17: Some weeks ago, we removed this patch for safety, but this should exist.
This action should be run when the commit has the expected substrings, but also when:
- A PR gets merged to main, to deploy things to the **testnet** network.
- When a release event happens, so the **mainnet** job is run.

.github/workflows/deploy.yml
```diff
@@ -78,7 +78,7 @@ jobs:
             for i in 1 2 3 4; do [ $(curl --insecure --silent https://127.0.0.1/api/v1/version | jq .git_commit) = "\"${{ steps.prep.outputs.sha_short}}\"" ] && break || (if [ $i = 4 ]; then exit -1; else sleep 5; fi ) done 
             echo "All healthy!"
 
-      - if: github.event_name == 'release' || github.ref == 'refs/heads/main' || contains(github.event.head_commit.message, '[mainnetdeploy]')
+      - if: github.event_name == 'release' || contains(github.event.head_commit.message, '[mainnetdeploy]')
```
jsign, line 81: This was incorrect since we don't want to ever deploy to mainnet when merging to main.
Deploying to mainnet should only happen on a release, or a correctly created commit message.

.github/workflows/deploy.yml
```diff
@@ -55,7 +55,7 @@ jobs:
             for i in 1 2 3 4; do [ $(curl --insecure --silent https://127.0.0.1/api/v1/version | jq .git_commit) = "\"${{ steps.prep.outputs.sha_short}}\"" ] && break || (if [ $i = 4 ]; then exit -1; else sleep 5; fi ) done 
             echo "All healthy!"
 
-      - if: github.ref == 'refs/heads/main'|| contains(github.event.head_commit.message, '[testnetdeploy]')
+      - if: github.ref == 'refs/heads/main' || contains(github.event.head_commit.message, '[testnetdeploy]')
```
jsign, line 58: No change here, just commenting that we deploy to testnet when merging to main.

brunocalza (approved):

PR #438, opened 2023-01-18
adds missing caret to regex

brunocalza:
# Summary

The caret was missing in the regex that validates the table's name

# Context

The regex must match the full table name. The same rule is used in the parser [helper](https://github.com/tablelandnetwork/go-sqlparser/blob/main/helpers.go#L87).


jsign (approved):
:+1:

PR #440, opened 2023-01-20
Healthbot fixes and re-enabling & Docker images updating

jsign:
# Summary

This PR:
- Fixes some stuff in the new healthbot that sends the transactions directly instead of relying on validator relaying.
- Updates some base Docker images to not fall behind in having everything up to date at the infra level.

# Context

NA

# Implementation overview

NA

NA


.github/workflows/deploy.yml
```diff
@@ -70,7 +70,7 @@ jobs:
             git fetch && git checkout -f ${{ github.sha }}
             cd docker
 
-            make testnet-up
+            HEALTHBOT_ENABLED=true make testnet-up
```
jsign, line 73: Re -enable healthbot in testnet now that we stopped the old one.

cmd/healthbot/main.go
```diff
@@ -49,7 +49,8 @@ func main() {
 			log.Fatal().Int("chain_id", chainCfg.ChainID).Msg("the chain id isn't supported in the Tableland client")
 		}
 
-		client, err := clientV1.NewClient(ctx, wallet, clientV1.NewClientChain(chain))
+		client, err := clientV1.NewClient(
+			ctx, wallet, clientV1.NewClientChain(chain), clientV1.NewClientAlchemyAPIKey(chainCfg.AlchemyAPIKey))
```
jsign, line 53: This is how I noticed the missing `AlchemyAPIKey` above. 

I think we can improve the UX in this client API by thinking of not making the API Key being an optional.
Despite being an optional is nice to have different flavors of "Infura", "Alchemy", etc, it's isn't correct in the sense that it isn't "optional". After you run it, it will break since it can't connect to any Ethereum API provider.

jsign: Created https://github.com/tablelandnetwork/go-tableland/issues/441 to discuss.

brunocalza (approved):

PR #442, opened 2023-01-20
dockerignore: avoid excluding files that are part of the repo

jsign:
# Summary

This PR changes the `.dockerignore` file that removes some files that are part of the repo when building the Docker images. This causes some problems for the "dirtyness" of the git repo in the Docker scratchpad, which ultimately makes the reported git version by te validator to always be `*-dirty`.

Tested with a beta release, and [it's working correctly now](https://tableland.network/api/v1/version).

# Context

NA.

# Implementation overview

NA.

NA

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,1 +1,6 @@
-docker
+docker/**/*.db*
+docker/**/backups
+docker/deployed/**/prometheus
+docker/deployed/**/grafana/data/**
+!docker/deployed/**/grafana/data/.gitkeep
+


.dockerignore
```diff
@@ -1 +1,6 @@
-docker
\ No newline at end of file
+docker/**/*.db*
+docker/**/backups
+docker/deployed/**/prometheus
+docker/deployed/**/grafana/data/**
+!docker/deployed/**/grafana/data/.gitkeep
```
jsign: The previous rule was too aggressive. We want to do a mimicking of the `.gitignore`, with an extra twist of allowing the `.gitkeep` files.

brunocalza (approved):

PR #455, opened 2023-02-06
[wip] Adds a  Merkle Tree library and "snapshotting" of tables' rows

brunocalza:
# Summary

This PR implements the architecture necessary to provide **Inclusion Proofs** in Tableland. 

- It relates to the GH project [Brute force Merkle Tree of all tables (experimental)](https://github.com/orgs/tablelandnetwork/projects/40/views/1).
- An in-depth discussion about the architecture and design decisions can be found in an internal Notion [doc](https://www.notion.so/The-Design-of-Tableland-Inclusion-Proofs-dc4a56a6e688415ea59dfa9929090edc?d=f74a43dbd45c40f5ba1f4bfed7c2e2c8#7473fcb54d5d4413b340b0c31eece79d).

# Context

This can be potentially a very big PR. Not really sure how it will fully unfold. But as is, we have fully working parts. And others to come next, as described in the next section.

# Implementation overview

Currently, we have the following parts implemented and ready for review:

- Merkle Tree library
   This is an independent package that implements a simple in-memory Merkle tree. It's currently not used by any code. But it will be used by Merkle Root publisher
- "Snapshotting" of tables' rows
    For every X block, we snapshot all rows for all tables into a stream of bytes row in a new table. This is described [here](https://www.notion.so/textile/The-Design-of-Tableland-Inclusion-Proofs-dc4a56a6e688415ea59dfa9929090edc#2eb1c961a3344fe9a1f85413e0d0bd13). 

Parts to come:
- #375 
  This part will fetch rows from '"Snapshotting" of tables' rows', build the Merkle Tree and publish the root to a smart contract
- #374 

# Todos

- [ ] Remove calculate hash implementation

# Notes
- It is being run in staging
- The Merkle Tree proof implementation was tested against [OpenZeppelin](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/260e082ed10e86e5870c4e5859750a8271eeb2b9/contracts/utils/cryptography/MerkleProof.sol) implementation  (this is probably an end-to-end test worth adding later)

# Checklist

- [ ]  Are changes backward compatible with existing SDKs, or is there a plan to manage it correctly?
- [ ]  Are code changes optimized for future code readers, commenting on problematic areas to understand (if any)?
- [ ]  Future-self question: Did you avoid unjustified/unnecessary complexity to achieve the goal?


pkg/eventprocessor/impl/executor/impl/blockscope.go
```diff
@@ -215,6 +216,84 @@ func (bs *blockScope) StateHash(ctx context.Context, chainID tableland.ChainID)
...
+		rowHash := sha3.New256()
+		for _, col := range rawBuffer {
+			rowHash.Write(col)
+		}
+
+		leaves = append(leaves, rowHash.Sum(nil)...)
+	}
+
+	if _, err := bs.txn.ExecContext(ctx,
+		"INSERT INTO system_tree_leaves (prefix, chain_id, table_id, block_number, leaves) VALUES (?1, ?2, ?3, ?4, ?5)",
```
brunocalza, line 284: Here's the "snapshotting". We're copying a table into an entire row as a stream of bytes. The information will be picked up by a background process. 

Not really sure how that will fully work, but one approach use this table a queue. So the background process pick rows form here, process, then deletes the row. With that approach the process would have to keep a copy into another table or database. Why? Because it needs to generate proofs from this data at any given time. 

Another approach is to put a flag in this table to indicate that it was processed. We could prune older rows. 

In my mind what sounds like a good idea is the first delete approach combined with the serialization of the merkle tree. Here's how it would work: the background process pick a row to process, builds a tree, serialize the tree into another table (could even be another database, maybe `trees.db`. Another db would make it harder to make it transactional), then published the root to the registry. So we would have the entire tree ready in the db, when the request for a proof comes.

pkg/merkletree/tree.go
```diff
@@ -0,0 +1,173 @@
...
+import (
+	"bytes"
+	"errors"
+	"sort"
+
+	"github.com/ethereum/go-ethereum/crypto"
+)
+
+// DefaultHashFunc is the default hash function in case none is passed.
+var DefaultHashFunc = crypto.Keccak256
```
brunocalza, line 12: const is better

pkg/merkletree/tree.go
```diff
@@ -0,0 +1,173 @@
...
+	parent, left, right *Node
+	hash                []byte
+}
+
+func (n *Node) isLeaf() bool {
+	return n.left == nil && n.right == nil
+}
+
+// NewTree builds a new Merkle Tree.
+func NewTree(leaves [][]byte, hashFunc func(...[]byte) []byte) (*MerkleTree, error) {
```
brunocalza, line 33: You can pass the hash function. It helps with testing and gives clients the freedom to choose. I picked the signature `func(...[]byte) []byte` because the `hash.Hash` interface is quite cumbersome. But maybe will change this in the future because in some situation the size of the hash is needed, and this signature does not provided that.

pkg/merkletree/tree.go
```diff
@@ -0,0 +1,173 @@
...
+	if bytes.Compare(node.left.hash, node.right.hash) > 0 {
+		return t.hashFunc(t.verify(node.right), t.verify(node.left))
+	}
+
+	return t.hashFunc(t.verify(node.left), t.verify(node.right))
+}
+
+// GetProof gets the proof for a particular content.
+func (t *MerkleTree) GetProof(content []byte) [][]byte {
+	index, found := sort.Find(len(t.leaves), func(i int) int {
```
brunocalza, line 124: executes a binary search

pkg/merkletree/tree.go
```diff
@@ -0,0 +1,173 @@
+package merkletree
+
+import (
```
brunocalza, line 3: Here's the implementation of the Merkle Tree. The to most important design decisions are describe in the Readme ([link](https://github.com/tablelandnetwork/go-tableland/pull/455/files#diff-8b60b48f44e6be8099f3885f4dc07f54dd625fe46d22ece1538fd780c907aecaR5-R8)). 

I think the implementation is very straightforward and the comments in the code should be enough

pkg/merkletree/tree_test.go
```diff
@@ -0,0 +1,357 @@
...
+	t.Run("wrong proof", func(t *testing.T) {
+		t.Parallel()
+		root := []byte("036")
+		proof := [][]byte{[]byte("001"), []byte("007"), []byte("026")}
+		leaf := []byte("001")
+		require.False(t, VerifyProof(root, proof, leaf, mockHashFunc))
+	})
+}
+
+func TestProperties(t *testing.T) {
```
brunocalza, line 177: Experimenting with the `testing/quick` library to test properties of the tree and catch problems. It was quite good actually, found 2 bugs in my first implementation.

pkg/merkletree/tree_test.go
```diff
@@ -0,0 +1,357 @@
...
+			h := sha3.Sum512(b[0])
+			return h[:]
+		},
+		func(b ...[]byte) []byte {
+			h := sha1.Sum(b[0])
+			return h[:]
+		},
+	}
+
+	t.Run("tree holds merkle tree property", func(t *testing.T) {
```
brunocalza, line 209: this tests if the tree holds the merkle tree property: the hashing of the leaves leads to the merkle root

pkg/merkletree/tree_test.go
```diff
@@ -0,0 +1,357 @@
...
+					return false
+				}
+
+				return tree.VerifyTree()
+			}
+			require.NoError(t, quick.Check(property, nil))
+		}
+	})
+
+	t.Run("leaves are always sorted", func(t *testing.T) {
```
brunocalza, line 229: This test that we did not messed up with the order along the way

pkg/merkletree/tree_test.go
```diff
@@ -0,0 +1,357 @@
...
+
+				return sort.SliceIsSorted(tree.leaves, func(i, j int) bool {
+					return bytes.Compare(tree.leaves[i].hash, tree.leaves[j].hash) == -1
+				})
+			}
+			require.NoError(t, quick.Check(property, nil))
+		}
+	})
+
+	t.Run("height of the tree is correct", func(t *testing.T) {
```
brunocalza, line 251: This tests that the height property of the tree

pkg/merkletree/tree_test.go
```diff
@@ -0,0 +1,357 @@
...
+				}
+
+				return len(tree.leaves) == len(leaves)+1 &&
+					bytes.Equal(tree.leaves[len(tree.leaves)-1].hash, tree.leaves[len(tree.leaves)-2].hash)
+			}
+			require.NoError(t, quick.Check(property, nil))
+		}
+	})
+
+	t.Run("every leaf proof is correctly verifiable", func(t *testing.T) {
```
brunocalza, line 290: This tests that every leaf can be correctly verified by getting the proof of each one and verifying

pkg/eventprocessor/impl/executor/impl/blockscope.go
```diff
@@ -215,6 +216,84 @@ func (bs *blockScope) StateHash(ctx context.Context, chainID tableland.ChainID)
 	return executor.NewStateHash(chainID, bs.scopeVars.BlockNumber, hash), nil
 }
 
+func (bs *blockScope) CalculateTreeLeaves(ctx context.Context, chainID tableland.ChainID) error {
```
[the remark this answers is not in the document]
brunocalza: Also, this implementation excludes system tables. That means we are losing the capability to compare a full database state. Maybe only for those tables, we can use the current state hash solution.

pkg/eventprocessor/impl/executor/impl/blockscope.go
```diff
@@ -215,6 +216,84 @@ func (bs *blockScope) StateHash(ctx context.Context, chainID tableland.ChainID)
...
+		if err := tableRows.Scan(scanCallArgs...); err != nil {
+			return fmt.Errorf("table row scan: %s", err)
+		}
+
+		rowHash := sha3.New256()
+		for _, col := range rawBuffer {
+			rowHash.Write(col)
+		}
+
+		leaves = append(leaves, rowHash.Sum(nil)...)
```
jsign, line 280: It maybe worth deduping here? The hash will be the same if two rows have the same column values. That would make the underlying tree have duplicate leaves next to each other and make the tree bigger unnecessarily.

jsign (dismissed):
LGTM!

I left some comments that I think you might want to take a look at.

pkg/eventprocessor/impl/executor/impl/blockscope.go
```diff
@@ -215,6 +216,84 @@ func (bs *blockScope) StateHash(ctx context.Context, chainID tableland.ChainID)
...
+		scanCallArgs[i] = &rawBuffer[i]
+	}
+
+	leaves := []byte{}
+	for tableRows.Next() {
+		if err := tableRows.Scan(scanCallArgs...); err != nil {
+			return fmt.Errorf("table row scan: %s", err)
+		}
+
+		rowHash := sha3.New256()
```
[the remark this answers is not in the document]
brunocalza: oh right, I always go for `sha` and forget about if the cryptographic properties are needed. good callout. any suggestions on standard go implementations? maybe crc32?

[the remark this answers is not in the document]
jsign: `fnv` is pretty fast and [supported in stdlib](https://pkg.go.dev/hash/fnv). It can be an option.

pkg/merkletree/tree.go
brunocalza: That comment was more a not for myself 😅

It's intentional because clients might need to know what hash was used to generate the leaves, so they can use the same. Ideally, that value would be a const, because you don't want clients messing with that. I'll probably switch to `hash.Hash`.

pkg/merkletree/tree.go
```diff
@@ -0,0 +1,173 @@
...
+	}
+
+	tree.buildTree(leaves)
+	return tree, nil
+}
+
+func (t *MerkleTree) buildTree(leaves [][]byte) {
+	t.leaves = make([]*Node, len(leaves))
+	for i, leaf := range leaves {
+		t.leaves[i] = &Node{
```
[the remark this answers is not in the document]
brunocalza: yeahh, good catch

pkg/merkletree/tree.go
jsign: Sounds good, but I think its fine to think about `pkg/merkletree` just as an algorithm implementation.
If `Keccak256` is relevant for a higher-level protocol, that's fine but can be considered relevant in an outer context.

But... biasing this package towards the protocol sounds fine too, it can help people importing `go-tableland` to use this package directly without "any further" consideration.

pkg/merkletree/tree.go
```diff
@@ -0,0 +1,173 @@
...
+	}
+
+	t.buildInternalNodes(parentNodes)
+}
+
+// VerifyTree calculates the merkle root again by traversing the tree and verify if it's the same it holds.
+func (t *MerkleTree) VerifyTree() bool {
+	merkleRoot := t.verify(t.root)
+	return bytes.Equal(t.root.hash, merkleRoot)
+}
```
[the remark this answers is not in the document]
brunocalza: yeah, this is just a double-check for tests to make sure the tree was constructed correctly. could be unexported

pkg/eventprocessor/impl/executor/impl/blockscope.go
```diff
@@ -215,6 +216,84 @@ func (bs *blockScope) StateHash(ctx context.Context, chainID tableland.ChainID)
...
+		for _, col := range rawBuffer {
+			rowHash.Write(col)
```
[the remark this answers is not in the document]
brunocalza: From what I understood there it's that it does not follow any standard behavior, but it has a deterministic behavior: 
> it is the current habit of choice of SQLite specifically to return the columns in the order they are created in the internal structure, I assume which is simply because it happens to be the order in which they are parsed, which happens to be the order in which they appear in the CREATE statement

The problem is that they can change that behavior in a future version (because it's not standard) and also ALTER tables can mess up things (we don't support `ALTER TABLE`, at least yet). 

We could sort it but that implies that the client when producing the leaf for verification, would have to sort it too. Maybe that's ok? I think so.

PR #460, opened 2023-02-14
remove the legacy API and stop relaying smart contract calls

### LINKED ISSUE #360: [GOT-35] Drop JSON-RPC support
null

<sub>[GOT-35](https://linear.app/tableland/issue/GOT-35/drop-json-rpc-support)</sub>

joewagner:
# Summary

This is a draft, but I wanted to explore the implications of removing the relaying and JSON RPC endpoints.
I've created a build of this branch and used it locally in LocalTableland, in combination with the changes in https://github.com/tablelandnetwork/evm-tableland/issues/306, to run the tests for the SDK and everything is passing without any changes to the SDK.

In theory this fixes #360

# Context

TODO

# Implementation overview

TODO

TODO 

# Checklist

- [x]  Are changes backward compatible with existing SDKs, or is there a plan to manage it correctly? The SDK no longer makes use of the JSON-RPC endpoints and does not use the transaction relays anymore.
- [x]  Are changes covered by existing tests, or were new tests included? I still have at least one questions on the tests for rate limiting based on the siwe token.


docker/deployed/mainnet/api/config.json
```diff
@@ -54,7 +54,6 @@
         {
             "Name": "Ethereum Mainnet",
             "ChainID": 1,
-            "AllowTransactionRelay": false,
```
joewagner, line 57: I'm not entirely sure what the implications of changing these config files is.

internal/router/middlewares/contextkeys.go
```diff
@@ -4,10 +4,10 @@ package middlewares
 type ContextKey int
 
 const (
-	// ContextKeyAddress is used to store the address of the client for the incoming request.
```
joewagner, line 7: The term "address" in this comment was confusing to me at first. I believe it is the wallet address, not the IP address, since the IP address is found in ContextIPAddress.
As such we can remove it because the siwe was the only place the wallet address was available in the request context.

internal/router/middlewares/ratelim.go
```diff
@@ -36,18 +26,10 @@ type RateLimiterRouteConfig struct {
...
-// 3. If 2. isn't present, it will use the connection remote address.
+// 1. If found, use an existing X-Forwarded-For IP included by a load-balancer in the infrastructure.
+// 2. If 1. isn't present, it will use the connection remote address.
 func RateLimitController(cfg RateLimiterConfig) (mux.MiddlewareFunc, error) {
 	keyFunc := func(r *http.Request) (string, error) {
-		// Use a chain address if present.
-		address := r.Context().Value(ContextKeyAddress)
-		ctrlAddress, ok := address.(string)
-		if ok && ctrlAddress != "" {
-			return ctrlAddress, nil
```
joewagner, line 48: We no longer has the "chain-address" available, so we can't rate limit with it anymore.

internal/router/middlewares/ratelim.go
```diff
@@ -59,50 +41,9 @@ func RateLimitController(cfg RateLimiterConfig) (mux.MiddlewareFunc, error) {
 	if err != nil {
 		return nil, fmt.Errorf("creating default rate limiter: %s", err)
 	}
-	customRLs := make(map[string]*httplimit.Middleware, len(cfg.JSONRPCMethodLimits))
-	for route, routeCfg := range cfg.JSONRPCMethodLimits {
```
joewagner, line 63: the RPC rate limiting is gone now, and this means we only need to use the default rate limiter.

internal/router/middlewares/ratelim_test.go
```diff
@@ -197,36 +74,6 @@ func TestLimit1IP(t *testing.T) {
...
-			MaxRPI:   150,
-			Interval: time.Second,
-		},
-		JSONRPCRoute: "/rpc",
-	}
-	rlcm, err := RateLimitController(cfg)
-	require.NoError(t, err)
-	rlc := rlcm(dummyHandler{})
-
-	// Do 1000 requests as fast as we can with *different addresses*, and see that
```
joewagner, line 215: I'm assuming we are referring to wallet addresses here, since we are looking at `ContextKeyAddress`

internal/router/middlewares/ratelim_test.go
```diff
@@ -14,126 +12,6 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-func TestLimit1Addr(t *testing.T) {
```
joewagner, line 17: Definitely look close at this file.  It makes me uncomfortable to remove so many tests.  I _think_ this is ok because we simplified the rate limiting so much, but flagging this as a high risk of regression.

internal/router/middlewares/ratelim_test.go
```diff
@@ -246,7 +92,7 @@ func TestRateLim10IPs(t *testing.T) {
 	// we never get a 429 status response. The request per second being done is
 	// clearly more than 10 per second, but from different addresses which should be fine.
 	for i := 0; i < 1000; i++ {
-		ctx := context.WithValue(context.Background(), ContextKeyAddress, strconv.Itoa(i))
+		ctx := context.WithValue(context.Background(), ContextIPAddress, strconv.Itoa(i))
```
joewagner, line 95: Based on the test description, `TestRateLim10IPs`, and the comments, I think this test might not have actually been working how we thought.
The context contains `ContextKeyAddress`, which is a *wallet address*.  I think this test was supposed to be testing with *IP address*.  
Regardless of what it was doing, it's now testing with IP address.

internal/router/router.go
```diff
@@ -59,32 +46,6 @@ func ConfiguredRouter(
...
-	router.get("/chain/{chainId}/tables/structure/{hash}", ctrl.GetTablesByStructureHash, middlewares.WithLogging, middlewares.OtelHTTP("GetTablesByStructureHash"), middlewares.RESTChainID(supportedChainIDs), rateLim) // nolint
-	router.get("/schema/{table_name}", ctrl.GetSchemaByTableName, middlewares.WithLogging, middlewares.OtelHTTP("GetSchemaFromTableName"), rateLim)                                                                       // nolint
-
-	router.get("/query", ctrl.GetTableQuery, middlewares.WithLogging, middlewares.OtelHTTP("GetTableQuery"), rateLim) // nolint
-	router.get("/version", ctrl.Version, middlewares.WithLogging, middlewares.OtelHTTP("Version"), rateLim)           // nolint
-
-	// Health endpoint configuration.
-	router.get("/healthz", controllers.HealthHandler)
-	router.get("/health", controllers.HealthHandler)
-}
```
joewagner, line 86: I removed all the legacy routes here. Double check that we actually want to get rid of all of these?

internal/router/router.go
```diff
@@ -170,13 +131,6 @@ func (r *Router) get(uri string, f http.HandlerFunc, mid ...mux.MiddlewareFunc)
 	sub.Use(mid...)
 }
 
-// post creates a subroute on the specified URI that only accepts POST. You can provide specific middlewares.
```
joewagner, line 173: nre, having the Validator no longer accept POST requests feels, to me, more conceptually inline with the protocol.

internal/tableland/impl/mesa_instrumented.go
```diff
@@ -110,22 +94,6 @@ func (t *InstrumentedTablelandMesa) GetReceipt(
 	return ok, resp, err
 }
 
-// SetController allows users to the controller for a token id.
-func (t *InstrumentedTablelandMesa) SetController(
```
joewagner, line 114: nre, after reviewing all these changes I noticed that we don't seem to have been doing relays for locking a controller, even though the contract would have enabled this.

internal/router/middlewares/ratelim_test.go
brunocalza: You're right

internal/router/middlewares/ratelim_test.go
brunocalza: yeah, makes sense

internal/router/middlewares/ratelim_test.go
brunocalza: Actually, the [IP Address comes from the HTTP request](https://github.com/tablelandnetwork/go-tableland/pull/460/files#diff-3fbceac783d96c5c5e7e85ccaeee8243d07d4faffff54e19c89ecd5415288181R33) and [not from the context as we did with the wallet address](https://github.com/tablelandnetwork/go-tableland/pull/460/files#diff-3fbceac783d96c5c5e7e85ccaeee8243d07d4faffff54e19c89ecd5415288181L45). This line is not actually needed (maybe it's better to remove it) for this test and is probably a leftover from copying code from other tests. 

And just for completeness, we do store the IP address in the context. That's why the variable `ContextIPAddress` exits. But that's for a different purpose (logging, and collecting metrics), and not rate limiting.

brunocalza (commented):
Nice work @joewagner Overall it looks pretty good :) 

I have some comments for discussion before moving forward with this:
- Looks like there are people still using the Legacy JSON RPC API. See [logs](https://console.cloud.google.com/logs/query;query=resource.type%3D%22gce_instance%22%0Aresource.labels.instance_id%3D%225361159262168818743%22%0A%22call%20to%20legacy%20RPC%22%0A-jsonPayload.body%3D%22%7B%5C%22jsonrpc%5C%22:%5C%222.0%5C%22,%5C%22method%5C%22:%5C%22tableland_runReadQuery%5C%22,%5C%22id%5C%22:1,%5C%22params%5C%22:%5B%7B%5C%22statement%5C%22:%5C%22SELECT%20*%20FROM%20healthbot_5_1%3B%5C%22,%5C%22output%5C%22:%5C%22table%5C%22%7D%5D%7D%22;timeRange=2023-02-15T07:14:15.421Z%2F2023-02-15T14:14:15.421811Z;summaryFields=jsonPayload%252Ferror,jsonPayload%252Fchain_name,jsonPayload%252Fbody:false:32:beginning;cursorTimestamp=2023-02-15T12:01:38.887109539Z?organizationId=733293853667&project=textile-310716). I can't remember if we gave them a deprecation notice and a period to migrate. I see that people are using the legacy API only for reads, and are not using the `tableland_relayWriteQuery`. That leaves us with two options:
  - Give a deprecation notice, wait the period, then merge the PR. This is the slow path.
  - Give a deprecation notice, still keep the legacy API for reads and only stop the relay. That would mean, breaking this work into two PRs. The first one we simply disable `AllowTransactionRelay` in the `config.json` for all chains. Then, we remove the legacy API. 
- You can search for `TODO(json-rpc)` in the code base and you'll find more stuff to remove reg the legacy API haha
- It would be good to add the `bytecode.bin` and `abi.json` of the new contract inside `pkg/tables/impl/ethereum` and run the tests to see if there's a test that is "relaying". That should be fixed

joewagner:
@brunocalza Thanks for looking this over!  
Splitting this into 2 parts sounds good to me.  I'll split this up and address all of your comments in their respective PRs

joewagner:
@brunocalza I opened PR #461 to handle disabling the relaying as a first step.  It is a pretty small change since, as you mentioned, all that is really needed is updating the config files.  I'll work on cleaning up this PR and it can serve as a means of shutting off the RPC endpoints when the time comes.

docker/observability/grafana/provisioning/dashboards/validator-dashboard.json
```diff
@@ -1760,7 +1760,7 @@
           },
           "editorMode": "code",
           "exemplar": true,
-          "expr": "histogram_quantile(0.95, sum(rate(http_server_duration_bucket{http_server_name!=\"rpc\"}[5m])) by (http_server_name, le))",
+          "expr": "histogram_quantile(0.95, sum(rate(http_server_duration_bucket{}[5m])) by (http_server_name, le))",
```
joewagner, line 1763: @brunocalza I'm not sure about how this line and really all the changes to this file.  Is there a way to check if these changes are working how we want?

brunocalza (dismissed):

avichalp (approved):

PR #461, opened 2023-02-15
set AllowTransactionRelay to false in all configs

joewagner:
# Summary

This PR will turn off all transaction relaying and enable upgrading the registry contract to disallow contract owner calling functions on behalf of eoas. https://github.com/tablelandnetwork/evm-tableland/pull/307

# Context

This is a first step in shutting down and removing the legacy JSON-RPC API. #460 

# Implementation overview

The implementation is fairly simple.  The config files have all been changed so the `AllowTransactionRelay` value is false.  

Along with changing the config files, this somewhat redundantly forces all `ChainStack`s so`AllowTransactionRelay` value is false.  

# Checklist

### DIFF
--- a/cmd/api/main.go
+++ b/cmd/api/main.go
@@ -269,10 +269,11 @@ func createChainIDStack(
 		return chains.ChainStack{}, fmt.Errorf("starting event processor: %s", err)
 	}
 	return chains.ChainStack{
-		Store:                 systemStore,
-		Registry:              registry,
-		EventProcessor:        ep,
-		AllowTransactionRelay: config.AllowTransactionRelay,
+		Store:          systemStore,
+		Registry:       registry,
+		EventProcessor: ep,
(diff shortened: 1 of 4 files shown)


cmd/api/main.go
```diff
@@ -269,10 +269,11 @@ func createChainIDStack(
...
 	return chains.ChainStack{
-		Store:                 systemStore,
-		Registry:              registry,
-		EventProcessor:        ep,
-		AllowTransactionRelay: config.AllowTransactionRelay,
+		Store:          systemStore,
+		Registry:       registry,
+		EventProcessor: ep,
+		// TODO: we can remove the AllowTransactionRelay config property entirely in a future PR
+		AllowTransactionRelay: false,
```
joewagner, line 276: This is redundant since all of the config files have been changed, but I don't see any harm in doing this.

brunocalza (approved):

PR #465, opened 2023-02-17
upgrade the actions/setup-go action

joewagner:
# Summary

I noticed that the binaries.yml workflow failed with a message saying "Node.js 12 actions are deprecated. Please update the following actions to use Node.js 16: actions/setup-go@v2." https://github.com/tablelandnetwork/go-tableland/actions/runs/4204737716

I'm not sure if upgrading the setup-go action to v3 will fix the issue that caused the binaries build to fail, but seems like a good thing to do anyway. 

# Context

n/a

# Implementation overview

n/a

n/a

# Checklist

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/textileio/go-tableland
 
-go 1.19
+go 1.18
 
 require (
 	cloud.google.com/go/bigquery v1.45.0
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/textileio/go-tableland
 
-go 1.18
+go 1.19
 
 require (
 	cloud.google.com/go/bigquery v1.45.0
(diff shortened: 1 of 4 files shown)


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/textileio/go-tableland
 
-go 1.18
+go 1.19
```
joewagner, line 3: This _might_ fix the issue with building the binaries, but I don't have a good way to test.

brunocalza (commented):
This PR does not fix the problem, see https://github.com/tablelandnetwork/go-tableland/pull/467 but I think it's ok to keep it. Upgrading `actions/setup-go` is a good idea. There's no need to update the `go.mod`, I'll do it in another PR.

brunocalza (approved):
👍

PR #470, opened 2023-02-24
otel: updates library

brunocalza:
# Summary

This PR updates the OTEL dependencies to the latest version.

It also updates
- `github.com/rs/zerolog`
- `golang.org/x/crypto`
- `golang.org/x/net`

# Context

n/a

# Implementation overview

Changes we had to make based on their [Release Notes](https://github.com/open-telemetry/opentelemetry-go/releases): 

- The instrument kind names in `go.opentelemetry.io/otel/sdk/metric` are updated to match the API
  - `InstrumentKindSyncCounter` is renamed to `InstrumentKindCounter`
  - `InstrumentKindSyncUpDownCounter` is renamed to `InstrumentKindUpDownCounter`
  - `InstrumentKindSyncHistogram` is renamed to `InstrumentKindHistogram`
  - `InstrumentKindAsyncCounter` is renamed to` InstrumentKindObservableCounter`
  - `InstrumentKindAsyncUpDownCounter` is renamed to `InstrumentKindObservableUpDownCounter`
  - `InstrumentKindAsyncGauge` is renamed to `InstrumentKindObservableGauge`
- The `RegisterCallback` method of the Meter in `go.opentelemetry.io/otel/metric` changed.
  - The named `Callback` replaces the inline function parameter.
  - `Callback` is required to return an error.
  - `Callback` accepts the added `Observer` parameter added. This new parameter is used by `Callback` implementations to observe values for asynchronous instruments instead of calling the `Observe` method of the instrument directly. 
  - The slice of `instrument.Asynchronous` is now passed as a variadic argument. 
- Deprecated
  - The `go.opentelemetry.io/otel/metric/instrument/asyncfloat64` package is deprecated. Use the instruments from `go.opentelemetry.io/otel/metric/instrument` instead.
  - The `go.opentelemetry.io/otel/metric/instrument/asyncint64` package is deprecated. Use the instruments from `go.opentelemetry.io/otel/metric/instrument` instead.
  - The `go.opentelemetry.io/otel/metric/instrument/syncfloat64` package is deprecated. Use the instruments from `go.opentelemetry.io/otel/metric/instrument` instead. 
  - The `go.opentelemetry.io/otel/metric/instrument/syncint64` package is deprecated. Use the instruments from `go.opentelemetry.io/otel/metric/instrument` instead.


cmd/healthbot/counterprobe/counterprobe.go
```diff
@@ -11,7 +11,7 @@ import (
 	logger "github.com/rs/zerolog/log"
 	clientV1 "github.com/textileio/go-tableland/pkg/client/v1"
 	"go.opentelemetry.io/otel/attribute"
-	"go.opentelemetry.io/otel/metric/instrument/syncint64"
```
brunocalza, line 14: There's no `go.opentelemetry.io/otel/metric/instrument/syncint64` package now. All instruments can be found in `go.opentelemetry.io/otel/metric/instrument`

cmd/healthbot/counterprobe/counterprobe.go
```diff
@@ -33,7 +33,7 @@ type CounterProbe struct {
 	mLastCounterValue    int64
 	mLastCheck           time.Time
 	mLastSuccessfulCheck time.Time
-	mLatencyHist         syncint64.Histogram
+	mLatencyHist         instrument.Int64Histogram
```
brunocalza, line 36: You'll this change across a bunch of files. This is the new Histogram type, you'll see that for all other types. For async types you'll see the keyword `Observable` in the middle of the type definition. [See](https://github.com/tablelandnetwork/go-tableland/pull/470/files#diff-7977292631e4d2d29d9b3eb030d818884abd90c3f5766170a7a5af6fc49c28a5R18) how instruments are created.

cmd/healthbot/counterprobe/metrics.go
```diff
@@ -14,40 +15,41 @@ func (cp *CounterProbe) initMetrics(chainName string) error {
...
 
-	mCounterValue, err := meter.AsyncInt64().Gauge(metricPrefix + ".counter_value")
+	mCounterValue, err := meter.Int64ObservableGauge(metricPrefix + ".counter_value")
 	if err != nil {
 		return fmt.Errorf("registering counter value gauge: %s", err)
 	}
 
 	instruments := []instrument.Asynchronous{mLastCheck, mLastSuccessfulCheck, mCounterValue}
-	if err := meter.RegisterCallback(instruments, func(ctx context.Context) {
+	if _, err := meter.RegisterCallback(func(ctx context.Context, o metric.Observer) error {
```
brunocalza, line 40: This is another change that happened. The signature of the `RegisterCallback`. You'll see a lot of this change.

note: we don't need the return value. That value can be used to Unregister a callback, which is something we don't do.

go.mod
```diff
@@ -121,4 +121,4 @@ require (
 )
 
 // add status code to HTTP request metrics, see https://github.com/open-telemetry/opentelemetry-go-contrib/pull/771
-replace go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp => github.com/textileio/opentelemetry-go-contrib/instrumentation/net/http/otelhttp v0.26.1-0.20220407173812-deccbe316725
+replace go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp => github.com/textileio/opentelemetry-go-contrib/instrumentation/net/http/otelhttp v0.26.1-0.20230224190607-b4c579e82b23
```
brunocalza, line 124: I had to update this library that we maintain as well. Here's the [patch](https://github.com/textileio/opentelemetry-go-contrib/commits/v1.14.0-patched).

cmd/healthbot/counterprobe/metrics.go
avichalp: makes sense.

avichalp: why is mutex required in every callback?

avichalp (approved):
Looks good to me 💯 🚀

cmd/healthbot/counterprobe/metrics.go
brunocalza: Good question! TBH I think all of them should have, but probably the ones that have are the ones we had problems with data race.

PR #493, opened 2023-03-20
Add arbitrum nova config

avichalp:
# Summary

This PR integrates the validator with the Arbitrum Nova network. 

# Context

Arbitrum Nova is a [Arbitrum public chain](https://developer.arbitrum.io/public-chains) built on top of the [AnyTrust](https://developer.arbitrum.io/inside-anytrust). It is cheaper than other rollups because it doesn't rely on the L1 chain for data availability. Instead, the DA is guaranteed by a committee. It means that in practice tableland's create and update queries could be performed for ~ $0.001 per query. It makes the tableland network with Arbitrum highly affordable for use cases that have a high write volume.

# Implementation overview

This PR simply adds a new mainnet config. The config includes an endpoint to fetch events from the Arbirtum Nova blockchain. We will use a `quicknode` endpoint for this purpose. The `eventfeed` connects with the Arbitrum Nova chain using the endpoint. And replays the queries on the SQLite database.

_Here are some new tables that are created on Nova (chain id: 42170) after syncing._

Screenshot 2023-03-21 at 10 43 38 PM

# Checklist
- [x] Add required env vars on the vm

### DIFF
--- a/docker/deployed/mainnet/api/config.json
+++ b/docker/deployed/mainnet/api/config.json
@@ -108,20 +108,20 @@
             "HashCalculationStep": 450
         },
         {
-            "Name": "Polygon Mainnet",
-            "ChainID": 137,
+            "Name": "Arbitrum Nova Mainnet",
+            "ChainID": 42170,
             "AllowTransactionRelay": false,
             "Registry": {
-                "EthEndpoint": "wss://polygon-mainnet.g.alchemy.com/v2/${VALIDATOR_ALCHEMY_POLYGON_MAINNET_API_KEY}",
-                "ContractAddress": "0x5c4e6A9e5C1e1BF445A062006faF19EA6c49aFeA"
+                "EthEndpoint": "https://distinguished-burned-cherry.nova-mainnet.discover.quiknode.pro/${VALIDATOR_QUICKNODE_ARBITRUM_NOVA_MAINNET_API_KEY}",
+                "ContractAddress": "0x4dad9cbBeDF7934B20641F135bb1202c123AbBD5"
             },
             "Signer": {
-                "PrivateKey": "${VALIDATOR_POLYGON_MAINNET_SIGNER_PRIVATE_KEY}"
+                "PrivateKey": "${VALIDATOR_ARBITRUM_NOVA_MAINNET_SIGNER_PRIVATE_KEY}"
             },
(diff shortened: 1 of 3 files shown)


docker/deployed/mainnet/api/config.json
```diff
@@ -108,32 +108,32 @@
             "HashCalculationStep": 450
         },
         {
-            "Name": "Polygon Mainnet",
```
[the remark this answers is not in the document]
avichalp: Yes, thanks good catch. I did it for testing out initially; will add it back in upcoming commits.

pkg/eventprocessor/eventfeed/impl/eventfeed.go
```diff
@@ -165,7 +165,8 @@ func (ef *EventFeed) Start(
 				ef.log.Warn().Err(err).Msgf("filter logs from %d to %d", fromHeight, toHeight)
 				if strings.Contains(err.Error(), "read limit exceeded") ||
 					strings.Contains(err.Error(), "Log response size exceeded") ||
-					strings.Contains(err.Error(), "is greater than the limit") {
+					strings.Contains(err.Error(), "is greater than the limit") ||
+					strings.Contains(err.Error(), "eth_getLogs and eth_newFilter are limited to a 10,000 blocks range") {
 					ef.maxBlocksFetchSize = ef.maxBlocksFetchSize * 80 / 100
```
avichalp, line 170: Quicknode only allows block range no bigger than 10k. If the event feed hits the limit, this is the part where the block range will be reduced until the request succeeds.

docker/deployed/mainnet/api/config.json
```diff
@@ -107,6 +107,34 @@
             },
             "HashCalculationStep": 450
         },
+        {
+            "Name": "Arbitrum Nova Mainnet",
+            "ChainID": 42170,
+            "AllowTransactionRelay": false,
```
brunocalza, line 113: This flag `AllowTransactionRelay` is being removed in the Drop JSON-RPC work. I think you can remove it, to avoid conflicts later

brunocalza (dismissed):
There's a contract address mismatch between the client and the validator. Other than that, looks good :)

brunocalza (approved):

PR #495, opened 2023-03-22
refactor and cleaning after rpc removal 

brunocalza:
# Summary

This work builds on top of https://github.com/tablelandnetwork/go-tableland/pull/460 and it is part of the Drop JSON-RPC project. 

# Context

This PR removes more stuff that was somewhat related to the RPC architecture and is not needed anymore. Besides that, it does some code refactoring and cleaning. 

# Implementation overview & details

At a high-level here is an explanation of what was done:

## Transferred the `mesa` code and deleted the file

The tableland package is not making sense anymore. The Tableland interface was removed and transferred to the Gateway interface. And the mesa implementation was moved to the Gateway implementation. Little by little, we can remove all code from tableland package and put it into better places.

## Renamed `system` to `gateway` and simplified the `gateway` implementation

System was a poor name choice in the early days of Tablelad POC. 
System was because of all operations you could do in system data (data not from user). It turns out that is simply the Gateway.

In short, all logic that supports our REST API can be found here:

Screen Shot 2023-03-22 at 16 33 28

The gateway is passed to controllers responsible for HTTP handling. 

## Removes the `user` `sqlstore`

Again, the `user` & system separation was a poor abstraction created in the early days. Moved the user functionality to system `sqlstore`. In the future database refactor, we'll remove the `system` nomenclature. 

## Removes the following dashboards panels
  - Txn confirmation retries
Screen Shot 2023-03-22 at 16 36 49

  - Txn Gas price bumps
Screen Shot 2023-03-22 at 16 37 18

  - Pending txns
Screen Shot 2023-03-22 at 16 37 44

  - Wallet - Nonces
Screen Shot 2023-03-22 at 16 38 07

Those panels we're necessary to observe the txns relaying operation. We're not relaying anymore, so we don't need them.


cmd/api/main.go
```diff
@@ -216,18 +205,8 @@ func createChainIDStack(
...
-		config.ChainID,
-		scAddress,
-		wallet,
-		tracker,
-	)
-	if err != nil {
-		return chains.ChainStack{}, fmt.Errorf("failed to create ethereum client: %s", err)
-	}
 
-	acl := impl.NewACL(systemStore, registry)
```
brunocalza, line 230: I realized that this acl component was not using the `registry`, so I removed it

cmd/api/main.go
```diff
@@ -446,6 +424,15 @@ func createChainStacks(
 		chainStacks[chainCfg.ChainID] = chainStack
 	}
 
+	epMap := make(map[tableland.ChainID]eventprocessor.EventProcessor)
```
brunocalza, line 427: This logic is kind of ugly, but I could not think of something better. In the user store [we needed to pass the resolver](https://github.com/tablelandnetwork/go-tableland/pull/495/files#diff-ae22ad1e20b89eaefa8e09e4ed89443ac10fdc555b9767625f3866e3853100afL102). Since we transferred that functionality to the system store, we gotta do the same to the system store. But we have one system store per chain, that's why we got to do these loops, and set the resolver for each system store.

cmd/api/main.go
```diff
@@ -477,46 +464,34 @@ func createAPIServer(
 	httpConfig HTTPConfig,
 	gatewayConfig GatewayConfig,
 	parser parsing.SQLValidator,
-	userStore *user.UserStore,
 	chainStacks map[tableland.ChainID]chains.ChainStack,
 ) (moduleCloser, error) {
-	instrUserStore, err := sqlstoreimpl.NewInstrumentedUserStore(userStore)
```
brunocalza, line 483: no more user store

cmd/api/main.go
```diff
@@ -477,46 +464,34 @@ func createAPIServer(
...
-	if err != nil {
-		return nil, fmt.Errorf("creating instrumented user store: %s", err)
-	}
-
-	mesaService := impl.NewTablelandMesa(parser, instrUserStore, chainStacks)
```
brunocalza, line 488: no more mesa. it's all part of the gateway

internal/gateway/gateway.go
```diff
@@ -0,0 +1,225 @@
+package gateway
```
brunocalza, line 1: There's no new code here. It's just the merge of what we had in `mesa.go` and `system/sqlstore.go`

internal/gateway/gateway_instrumented.go
```diff
@@ -1,4 +1,4 @@
-package impl
```
brunocalza, line 1: just renaming going on in this file

internal/router/controllers/controller.go
```diff
@@ -30,15 +29,13 @@ type SQLRunner interface {
 
 // Controller defines the HTTP handlers for interacting with user tables.
 type Controller struct {
```
brunocalza, line 31: The controller had these two dependencies: `SQLRunner` (implemented by the old `mesa.go`) and `system.SystemService`. It's all `Gateway` now.

internal/router/controllers/controller.go
```diff
@@ -167,16 +164,8 @@ func (c *Controller) GetTable(rw http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	isAPIV1 := strings.HasPrefix(r.RequestURI, "/api/v1/tables")
```
brunocalza, line 170: removed this check because all api is APIV1 now

pkg/sqlstore/impl/system/store.go
```diff
@@ -79,6 +82,11 @@ func New(dbURI string, chainID tableland.ChainID) (*SystemStore, error) {
 	return systemStore, nil
 }
 
+// SetReadResolver sets the resolver for read queries.
+func (s *SystemStore) SetReadResolver(resolver sqlparser.ReadStatementResolver) {
```
brunocalza, line 86: Cannot pass this at the constructor level because of cyclical dependency between the event processor and the `SystemStore`, so i created a setter

pkg/sqlstore/impl/system/store.go
```diff
@@ -303,6 +311,33 @@ func (s *SystemStore) GetID(ctx context.Context) (string, error) {
 	return id, err
 }
 
+// Read executes a read statement on the db.
+func (s *SystemStore) Read(ctx context.Context, rq parsing.ReadStmt) (*tableland.TableData, error) {
```
brunocalza, line 315: This comes from the old user store

pkg/client/chains.go
```diff
@@ -102,11 +102,6 @@ var Chains = map[ChainID]Chain{
 	},
 }
 
-// CanRelayWrites returns whether Tableland validators will relay write requests.
-func (c Chain) CanRelayWrites() bool {
```
avichalp, line 106: 👍

internal/tableland/impl/mesa_test.go
avichalp, line 1: Do we want to rename this file since mesa is now gone?

avichalp: yeah, sounds good 👍

avichalp (approved):
Looks great, @brunocalza 👍

PR #497, opened 2023-03-28
Bumps the parser

brunocalza:
# Summary

Bumps the parser to the latest version. This version includes:
- [shorter string output for parsed statements](https://github.com/tablelandnetwork/go-sqlparser/pull/45)
- [fixes bug on upsert walk](https://github.com/tablelandnetwork/go-sqlparser/pull/49)

### DIFF
--- a/pkg/eventprocessor/impl/eventprocessor_replayhistory_test.go
+++ b/pkg/eventprocessor/impl/eventprocessor_replayhistory_test.go
@@ -35,14 +35,14 @@ func TestReplayProductionHistory(t *testing.T) {
 	}
 
 	expectedStateHashes := map[tableland.ChainID]string{
-		1:      "ce4f083e256d3458a329b6cd1ba7d8e93d9703b3",
-		5:      "08af7e7055f266ea75d3b06350c47a52cfa59843",
-		10:     "2bce2d62e7f5eeadc4736f470ca2959b871e5d91",
-		69:     "643af9ad784444242c6ef415727203941a720197",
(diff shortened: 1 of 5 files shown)


pkg/eventprocessor/impl/eventprocessor_replayhistory_test.go
```diff
@@ -35,14 +35,14 @@ func TestReplayProductionHistory(t *testing.T) {
 	}
 
 	expectedStateHashes := map[tableland.ChainID]string{
```
brunocalza, line 37: This version of the parser breaks the final hash state. We gotta update the hashes. Not a big deal.

sanderpick: out of curiosity, what are these hashed?

brunocalza: This is a test where we replay some historical events and calculate the final state of the database. Each chain has a final state hash. This test allows seeing if the change we made broke history and also allows us to see if our execution is deterministic. 

The new parser version breaks history because the statements are smaller (without unneeded spaces). Technically, that does not change the execution of events and user data should be the same. Still, some error messages that we store in receipts tables are affected by the size of the statement, thus, breaking history. Not a big deal though.

sanderpick:
Ah, remembering now. Thanks!

joewagner (approved):
LGMT!

PR #500, opened 2023-03-30
Replace old table external_url

joewagner:
This updates the external_url value of table metadata to use the new api endpoints, which is causing issues rendering the table nft on some of the market places.

### DIFF
--- a/internal/gateway/gateway.go
+++ b/internal/gateway/gateway.go
@@ -89,7 +89,7 @@ func (g *GatewayService) GetTableMetadata(ctx context.Context, id tables.TableID
 	chainID, store, err := g.getStore(ctx)
 	if err != nil {
 		return sqlstore.TableMetadata{
-			ExternalURL: fmt.Sprintf("%s/chain/%d/tables/%s", g.extURLPrefix, chainID, id),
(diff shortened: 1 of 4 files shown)


internal/gateway/gateway.go
```diff
@@ -89,7 +89,7 @@ func (g *GatewayService) GetTableMetadata(ctx context.Context, id tables.TableID
 	chainID, store, err := g.getStore(ctx)
 	if err != nil {
 		return sqlstore.TableMetadata{
-			ExternalURL: fmt.Sprintf("%s/chain/%d/tables/%s", g.extURLPrefix, chainID, id),
+			ExternalURL: fmt.Sprintf("%s/api/v1/tables/%d/%s", g.extURLPrefix, chainID, id),
```
avichalp, line 92: 👍

avichalp:
@joewagner LGTM. If I understand it right, we have clients that use the old external URL, causing rendering problems?

joewagner:
> @joewagner LGTM. If I understand it right, we have clients that use the old external URL, causing rendering problems?

Yeah, @asutula noticed the problem a few mintues ago.  I'll cc you in discord

avichalp (approved):

PR #501, opened 2023-03-31
Refactors how database is accessed

### LINKED ISSUE #502: [GOT-77] [GOT-28] Refactor the code to simplify the database access
<sub>From [SyncLinear.com](https://synclinear.com) | [GOT-28](https://linear.app/tableland/issue/GOT-28/refactor-the-code-to-simplify-the-database-access)</sub>

<sub>[GOT-77](https://linear.app/tableland/issue/GOT-77/[got-28]-refactor-the-code-to-simplify-the-database-access)</sub>

brunocalza:
# Summary

This PR refactors how we access the database in an attempt to make code simpler, and clearer and avoid some lock errors

The context is explained in the [Database Access Code Refactoring](https://www.notion.so/textile/Notes-on-how-database-is-accessed-refactoring-work-fa59cce9d85c4b178a60bf695dc57f44) document. This document should be read, the motivation well understood and a good mental model of the changes should be built before reviewing.

Closes #502


pkg/database/sqlite_db.go
```diff
@@ -0,0 +1,109 @@
...
+func OpenSerializable(path string, attributes ...attribute.KeyValue) (*SQLiteDB, error) {
+	attributes = append(attributes, attribute.String("type", "serializable"))
+	return Open(path, 1, attributes...)
+}
+
+// OpenConcurrent opens a SQLite database that allows multiple connections.
+// Should be used for reads.
+func OpenConcurrent(path string, attributes ...attribute.KeyValue) (*SQLiteDB, error) {
+	attributes = append(attributes, attribute.String("type", "concurrent"))
+	return Open(path, 0, attributes...)
```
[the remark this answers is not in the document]
avichalp: although this is supposed to be concurrent do we still want to have an upper cap on it in order to be protective of system resources?

tests/sqlite3.go
```diff
@@ -11,8 +11,9 @@ import (
 
 // Sqlite3URI returns a URI to spinup an in-memory Sqlite database.
 func Sqlite3URI(t *testing.T) string {
-	dbURI := "file::" + uuid.NewString() + ":?mode=memory&cache=shared&_foreign_keys=on"
+	dbURI := "file::" + uuid.NewString() + ":?mode=memory&cache=shared&_foreign_keys=on&_busy_timeout=5000"
```
avichalp, line 14: is the `busy_timeout` flag only relevant for the tests because we want to exit the long running test? or do we need the same flag in the usual flow too?

brunocalza: We're using that flag in the [usual flow ](https://github.com/tablelandnetwork/go-tableland/blob/main/cmd/api/main.go#L67) too. I just added it here to make the tests db and the real with equal flags.

pkg/database/sqlite_db.go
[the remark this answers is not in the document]
brunocalza: Good question. Probably, but don't know what that value could be. 

Also, not sure how much a cap would help. Suppose we have a cap of 500. That means at most 500 concurrent access can happen at the `*sql.DB` layer. But at the HTTP layer, there's no limit (only the rate limiter, but that's limited by IP address) as far as I'm concerned. That means the gateway could still be receiving much more than 500 HTTP requests which would probably affect the system resources. But those HTTP request would have a bigger wait time because now the `*sql.DB` locking them. I think my point is: having a cap at the database layer does not help much if you don't have a cap at the HTTP layer, or at least a way of refusing more HTTP connections if the database is "busy".

anyway, good question. gotta think more about this

internal/tableland/impl/acl.go
```diff
@@ -7,38 +7,93 @@ import (
...
+	return systemACL, nil
+}
+
+// SystemACL represents the system acl table.
+type SystemACL struct {
+	Controller string
+	ChainID    tableland.ChainID
+	TableID    tables.TableID
+	Privileges tableland.Privileges
+	// CreatedAt  time.Time
```
avichalp, line 97: why are we leaving these commented out?

pkg/nonce/impl/tracker_test.go
```diff
@@ -146,14 +149,15 @@ func TestInitialization(t *testing.T) {
 	wallet, err := wallet.NewWallet(hex.EncodeToString(crypto.FromECDSA(key)))
 	require.NoError(t, err)
 
-	sqlstore, err := system.New(url, tableland.ChainID(1337))
+	db, err := database.Open(url, 1)
+	// sqlstore, err := system.New(url, tableland.ChainID(1337))
```
avichalp, line 153: maybe worth removing this comment too, if not useful

internal/tableland/impl/acl.go
brunocalza: they are not really used and don't make too much sense in a decentralized world. gonna remove them

internal/tableland/impl/acl.go
```diff
@@ -7,38 +7,91 @@ import (
...
+
 	isAllowed, _ := aclRule.Privileges.CanExecute(op)
 	if !isAllowed {
 		return false, nil
 	}
 
 	return true, nil
 }
+
+func aclFromSQLtoDTO(acl db.SystemAcl) (SystemACL, error) {
```
avichalp, line 64: maybe it would be good to add a small docstring here stating what DTO means

brunocalza: sounds good

internal/tableland/impl/acl.go
```diff
@@ -7,38 +7,92 @@ import (
...
 	isAllowed, _ := aclRule.Privileges.CanExecute(op)
 	if !isAllowed {
 		return false, nil
 	}
 
 	return true, nil
 }
+
+// transforms the ACL data transfer object to ACL object model.
+func transformToObject(acl db.SystemAcl) (SystemACL, error) {
```
avichalp, line 65: 👍

avichalp (approved):
looks great!

avichalp (approved):

PR #504, opened 2023-04-04
Revert "Add FEVM support"

avichalp:
Reverts tablelandnetwork/go-tableland#498

After deploying the changes to `testnet` we ran into the [issue](https://github.com/filecoin-project/ref-fvm/issues/1350) of duplicate events in FVM. The `eventfeed` component doesn't expect duplicate events. 

Screenshot 2023-04-04 at 10 13 07 PM

We also have a `UNIQUE` constraint in the database. It throws an error if we try inserting events with the same transaction hash and event index.

Screenshot 2023-04-04 at 10 11 56 PM

The FVM issue mentioned above is currently a work in progress. However, in the meantime, to rectify the problem @brunocalza  and I are thinking about how we can deduplicate the events on our side. Meanwhile this revert PR will revert all FEVM related changes from testnet.


brunocalza (approved):

sanderpick:
hey guys, if you're seeing duplicate events... it might be useful to post the findings in fil-fvm-dev slack: https://filecoinproject.slack.com/archives/C029MT4PQB1. 

I suppose it could also be an issue with the provider.

PR #508, opened 2023-04-06
Changes the behavior of `block_num(<chain_id>)`

brunocalza:
# Summary

This PR changes the behavior of the custom function `block_num(<chain_id>)`.

# Context

That function was returning the latest block that contained a Tableland event, and it should return the latest block seen from the network provider's perspective. 

# Implementation overview

- It implements a new data structure called `SharedMemory` to make it easy to exchange data between different threads. In this particular case, we want to exchange the last block seen from each network thread and make it available to the gateway. This was already a component I have been discussing with @avichalp about adding. Can be useful for other stuff.
- The `SharedMemory` is passed to the `EventFeed` that has awareness of the blocks, and also passed to the Gateway's resolver, responsible for resolving the `block_num(<chain_id>)` function.


cmd/api/main.go
```diff
@@ -30,7 +30,7 @@ import (
 	"github.com/textileio/go-tableland/pkg/database"
 	"github.com/textileio/go-tableland/pkg/eventprocessor"
 	"github.com/textileio/go-tableland/pkg/eventprocessor/eventfeed"
-	"github.com/textileio/go-tableland/pkg/readstatementresolver"
```
brunocalza, line 33: I removed the `readstatementresolver` package. It was created to solve a cyclical dependency issue. Now the `readstatementresolver` is part of the `parsing` package.

cmd/api/main.go
```diff
@@ -93,10 +93,14 @@ func main() {
 		log.Fatal().Err(err).Msg("creating parser")
 	}
 
+	// Shared memory
```
brunocalza, line 96: Creates the shared memory object that will be shared with [`EventFeed`](https://github.com/tablelandnetwork/go-tableland/pull/508/files#diff-ae22ad1e20b89eaefa8e09e4ed89443ac10fdc555b9767625f3866e3853100afR247) and [Gateway's resolver](https://github.com/tablelandnetwork/go-tableland/pull/508/files#diff-ae22ad1e20b89eaefa8e09e4ed89443ac10fdc555b9767625f3866e3853100afR473)

pkg/client/v1/client_test.go
```diff
@@ -125,6 +125,25 @@ func TestGetTableByID(t *testing.T) {
 	})
 }
 
+func TestBlockNumber(t *testing.T) {
```
brunocalza, line 128: Created an e2e test for the block number in the client pkg. We have yet to find the best place for e2e tests. This package already has some, and it has some nice helpers that make it very quick to create an e2e test.

pkg/eventprocessor/eventfeed/impl/eventfeed.go
```diff
@@ -122,6 +128,7 @@ func (ef *EventFeed) Start(
 
 	// Listen for new blocks, and get new events.
 	for h := range chHeads {
+		ef.sm.SetLastSeenBlockNumber(ef.chainID, h.Number.Int64())
```
brunocalza, line 131: This is the place where the block number is being set in the shared memory object. It's right after we detected a new block header.

pkg/parsing/readstatementresolver.go
```diff
@@ -0,0 +1,21 @@
...
+}
+
+// NewReadStatementResolver creates a new ReadStatementResolver.
+func NewReadStatementResolver(sm *sharedmemory.SharedMemory) *ReadStatementResolver {
+	return &ReadStatementResolver{sm: sm}
+}
+
+// GetBlockNumber returns the block number for a given chain id.
+func (rqr *ReadStatementResolver) GetBlockNumber(chainID int64) (int64, bool) {
+	return rqr.sm.GetLastSeenBlockNumber(tableland.ChainID(chainID))
```
brunocalza, line 20: The resolver uses the shared memory object to get the value

pkg/eventprocessor/eventfeed/impl/eventfeed.go
```diff
@@ -146,6 +152,8 @@ func (ef *EventFeed) Start(
 				break
 			}
 
+			ef.sm.SetLastSeenBlockNumber(ef.chainID, toHeight)
```
brunocalza, line 155: This is the place where we're setting the last seen block number. It's right after we received a new block notification and we adjusted to account for reorgs. So, it's not exactly the last seen. it's the last seen minus our configuration. For Ethereum `mainnet`, it's the last seen minus 5.

avichalp (approved):
LGTM

brunocalza:
This was closed in favor of #527

PR #521, opened 2023-04-10
[GOT-12] Add GHA workflow with goreleaser

avichalp:
# Summary

This PR updates the binary generation workflow to use [goreleaser](https://goreleaser.com/) with [goreleaser-cross](https://github.com/goreleaser/goreleaser-cross) and produces a `darwin/arm64` binary.

# Context

The problem with the current setup is that it doesn't produce a binary for the new Apple Silicon architecture. We rely on GHA virtual machines to build these binaries. But GHA doesn't have a `darwin/arm64` VM support yet. We also rely on `CGO_ENABLED` flag because we use `go-sqlite3` along with its c bindings. It doesn't allow us to cross-compile the source code easily. 

Goreleaser is a useful tool to streamline the binary generation process. It automates the build process by using a simple config file and with minor changes in the GHA workflow. But it, too, cannot cross-compile source code that relies on c/c++ dependencies. 

Goreleaser-cross is a wrapper over Goreleaser that cross-compiles builds in a container and can build with c/c++ dependencies. It all supports all the  platfrom/arch pairs that we are currently interested in: `darwin/amd64`, `darwin/arm64`, `linux/amd64`, `linux/arm64`, `windows/amd64`. 

# Implementation overview

This PR adds the following: 

- A config file for `goreleaser`
- Updates the current binary generation workflow to use `goreleaser` and `goreleaser-cross`

[Here](https://github.com/avichalp/go-tableland/releases/tag/v1.0.1-beta-7) is an example of the generated binaries. It is used using the same branch but on a different fork of go-tableland: https://github.com/avichalp/go-tableland

# Checklist


.github/workflows/binaries.yml
```diff
@@ -3,43 +3,32 @@ on:
...
-            asset_name: api-windows-amd64.zip
-            compress_cmd: tar.exe -a -c -f
-            build_cmd: go build -o api ./cmd/api
-          - os: macos-latest
-            asset_name: api-darwin-amd64.tar.gz
-            compress_cmd: tar -czvf
-            build_cmd: go build -o api ./cmd/api
-    runs-on: ${{ matrix.os }}
+    runs-on: ubuntu-latest
+    container:
```
avichalp, line 11: [goreleaser-cross](https://github.com/goreleaser/goreleaser-cross) builds the binary in this container.

.github/workflows/binaries.yml
```diff
@@ -3,43 +3,32 @@ on:
...
         with:
           go-version: v1.19.x
-      - name: Build binary
-        run: ${{ matrix.build_cmd }}
-      - name: Pack output
-        run: ${{ matrix.compress_cmd }} ${{ matrix.asset_name }} api
-      - name: Upload binary
-        uses: svenstaro/upload-release-action@v2
+      - name: Release
+        uses: goreleaser/goreleaser-action@v4
```
avichalp, line 25: this step runs the `goreleaser release` to trigger binary generation and hence replaces the old binary generation process.

goreleaser.yml
avichalp, line 1: goreleaser config file according to which all the builds are generated.

.github/workflows/binaries.yml
```diff
@@ -3,43 +3,32 @@ on:
   release:
     types:
       - created
+permissions:
```
avichalp, line 6: gives permission to upload artifacts

sanderpick (approved):
Looks great!

Next time we do a major version bump, I think we should change the name of the binaries to something other than `api-`, which is pretty vague. How about `tableland-`?

avichalp:
thanks for the review @sanderpick!

> Next time we do a major version bump, I think we should change the name of the binaries to something other than api-, which is pretty vague. How about tableland-?

yes, this sounds good! maybe we can name it `tableland-api` or `tableland-validator`. would be good to think about the name in the meantime :)

PR #527, opened 2023-04-12
changes the behavior of block_num(<chain_id>) 

brunocalza:
# Summary

This PR changes the behavior of the custom function `block_num(<chain_id>)`.

# Context

That function was returning the latest block that contained a Tableland event, and it should return the latest block seen from the network provider's perspective. 

# Implementation overview

- It implements a new data structure called `SharedMemory` to make it easy to exchange data between different threads. In this particular case, we want to exchange the last block seen from each network thread and make it available to the gateway. This was already a component I have been discussing with @avichalp about adding. Can be useful for other stuff.
- The `SharedMemory` is passed to the `EventFeed` that has awareness of the blocks, and also passed to the Gateway's resolver, responsible for resolving the `block_num(<chain_id>)` function.


avichalp (approved):
looks good 👍

Issue #535, opened 2023-04-18
[GOT-101] [GOT-100] Validator release workflow doesn't connect to js-validator after upgrade to `goreleaser`


joewagner:
The update to use goreleaser to build binaries works well within go-tableland, but the automated js-validator pull requests aren't able to include the new builds because the binary name includes a shortened commit hash in the name.

Since js-validator doesn't know how to get the name of the zipped binary it can't download the file.

I'll submit a PR with a fix that uses the old naming scheme.  If having the commit hash in the name is important, we can explore alternatives.

<sub>From [SyncLinear.com](https://synclinear.com) | [GOT-100](https://linear.app/tableland/issue/GOT-100/validator-release-workflow-doesnt-connect-to-js-validator-after)</sub>

<sub>From [SyncLinear.com](https://synclinear.com) | [GOT-100](https://linear.app/tableland/issue/GOT-100/validator-release-workflow-doesnt-connect-to-js-validator-after)</sub>

sanderpick:
bummer. very odd that it only happens once in a while. i mean, looking at the code of the sync app i'm not surprised… it's spaghetti town. if it starts to happen w/ high freq i can look into it some more.

joewagner:
This should be fixed now, but we will find out when the release happens today.

joewagner:
fixed by #536

PR #536, opened 2023-04-18
remove commit hash from binary names

joewagner:
# Summary
This a fix for issue #535 
# Context
When we upgraded the binary build process to include builds for arm64 (i.e. apple silicon), the name template of the binary files was changed to include a short hash of the commit that the build is targeting.
This caused an issue in the build process of the [js-validator repo](https://github.com/tablelandnetwork/js-validator/blob/main/src/build.ts#L64).
The TL;DR; of the js-validator issue is that the script to download and unpack the go-tableland release doesn't have access to the short commit hash that now makes up part of the url where the binary is located.
# Implementation overview
This fix simply removes the commit hash from the file name.  If keeping the commit hash is important, we can close this PR and find a way for the js-validator build script to get the commit hash during the build.

----------------------

### DIFF
--- a/.github/workflows/binaries.yml
+++ b/.github/workflows/binaries.yml
@@ -16,7 +16,7 @@ jobs:
         with:
           submodules: 'true'
       - name: Fetch repo
-        run: git fetch --prune --unshallow      
+        run: git fetch --prune --unshallow
       - name: Install Go
         uses: actions/setup-go@v3
         with:
--- a/goreleaser.yml
+++ b/goreleaser.yml
@@ -91,14 +91,14 @@ archives:
       - api-darwin-arm64
       - api-linux-amd64
       - api-linux-arm64
-    name_template: "{{ .ProjectName }}-{{ .Os }}-{{ .Arch }}-{{ .ShortCommit }}"
+    name_template: "{{ .ProjectName }}-{{ .Os }}-{{ .Arch }}"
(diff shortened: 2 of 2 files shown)


.github/workflows/binaries.yml
```diff
@@ -16,7 +16,7 @@ jobs:
         with:
           submodules: 'true'
       - name: Fetch repo
-        run: git fetch --prune --unshallow      
+        run: git fetch --prune --unshallow
```
joewagner, line 19: Not related to this PR, just some cleanup

goreleaser.yml
```diff
@@ -91,14 +91,14 @@ archives:
       - api-darwin-arm64
       - api-linux-amd64
       - api-linux-arm64
-    name_template: "{{ .ProjectName }}-{{ .Os }}-{{ .Arch }}-{{ .ShortCommit }}"
+    name_template: "{{ .ProjectName }}-{{ .Os }}-{{ .Arch }}"
```
joewagner, line 94: Short and sweet, just remove the hash.

avichalp (approved):
looks good. the short hash doesn't serve any purpose. i just thought it would be good info to have in the name.

joewagner:
> looks good. the short hash doesn't serve any purpose. i just thought it would be good info to have in the name.

Awesome! that's an easy fix.

PR #540, opened 2023-04-20
Refactor provider selection in client

avichalp:
# Summary

Refactor `getContractBackend` function in the client package to make it easy to work with multiple RPC providers.

# Context
Earlier, we worked with only Infura and Alchemy but recently, we have added Anrk and Quicknode to use their RPC endpoints. It has made the existing code a bit unreadable and hard to maintain. This PR aims to simplify the implementation of the `getContractBackend` to make it easy to maintain.

# Implementation overview

This pull request refactors the `getContractBackend` function to simplify the selection of the Ethereum JSON-RPC provider by using a switch statement and introducing a new provider type. The previous implementation utilized multiple if-else conditions, making the code harder to read and maintain.

The changes in this PR include the following:

- Introducing a new provider type with name and apiKey fields.
Updating the config struct to use the new provider type instead of separate fields for each provider's API key and name.

- Removing the redundant checks for API keys and contract backend in the if-else conditions.

- Replacing the if-else conditions with a switch statement based on the config.provider value. Introducing a new variable rpcURL to store the URL template for the selected provider.

- Returning an error if no provider or ETH backend is specified in the configuration.

# Checklist

### DIFF
--- a/pkg/client/v1/client.go
+++ b/pkg/client/v1/client.go
@@ -32,9 +32,21 @@ type Client struct {
 	baseURL     *url.URL
 }
 
+// providerType can have possible value denoting Alchemy, Ankr, Infura etc.
+type providerType int
+
+const (
+	alchemy providerType = iota
+	infura
+	quickNode
+	ankr
+	local
+)
+
 type provider struct {
-	name   string
-	apiKey string
+	name         string
+	apiKey       string
+	providerType providerType
 }
 
 type config struct {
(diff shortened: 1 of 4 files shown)


asutula (approved):
Nice improvement.

pkg/client/v1/client.go
```diff
@@ -54,28 +56,28 @@ func NewClientChain(chain client.Chain) NewClientOption {
 // NewClientInfuraAPIKey specifies an Infura API to use when creating an EVM backend.
 func NewClientInfuraAPIKey(key string) NewClientOption {
 	return func(c *config) {
-		c.infuraAPIKey = key
+		c.provider = provider{name: "Infura", apiKey: key}
```
brunocalza, line 59: suggestion: instead of the hardcoded string, use an enum

brunocalza (approved):
looks good! left a comment for you consideration

brunocalza (dismissed):

brunocalza (approved):

PR #541, opened 2023-04-20
TAB-17: Update render endpoint

awmuncy:
# Summary

The Table's NFT renderer will be updating it's endpoints, which change both the structure of the endpoint, and the DNS. This pull request makes those changes so that the metadata of a table will point to the correct place. 

# Checklist


avichalp:
looks good to me. i am unsure if we want to make this PR against staging and merge it into staging for now. may be @joewagner can suggest

awmuncy:
> looks good to me. i am unsure if we want to make this PR against staging and merge it into staging for now. may be @joewagner can suggest

Good point; I've pointed it to staging.

avichalp (approved):
LGTM

PR #542, opened 2023-04-20
updates parser

brunocalza:
# Summary

Updates the parser 

# Context

When an identifier or alias had a dot, the stringification of the AST was producing an invalid statement.
See https://github.com/tablelandnetwork/go-sqlparser/pull/52

### DIFF
--- a/pkg/eventprocessor/impl/eventprocessor_replayhistory_test.go
+++ b/pkg/eventprocessor/impl/eventprocessor_replayhistory_test.go
@@ -37,12 +37,12 @@ func TestReplayProductionHistory(t *testing.T) {
 
 	expectedStateHashes := map[tableland.ChainID]string{
 		1:      "bce26781eed109b8aaae2d1f688c134831fdf061",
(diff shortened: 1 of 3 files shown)


pkg/eventprocessor/impl/eventprocessor_replayhistory_test.go
```diff
@@ -37,12 +37,12 @@ func TestReplayProductionHistory(t *testing.T) {
 
 	expectedStateHashes := map[tableland.ChainID]string{
```
brunocalza, line 38: Usually, when we change the parser we break history. In this case, we are now accepting statements that were not being accepted. This means that when we execute historical events we get a different state. That's why we need to update the test state hashes. Only testnets are being updated here, so not a big deal.

asutula (approved):
nice!

pkg/eventprocessor/impl/eventprocessor_replayhistory_test.go
avichalp: yes!

avichalp (approved):
LGTM

PR #549, opened 2023-04-28
fixes the response of receipt api call

### LINKED ISSUE #522: [GOT-92] The Validator API endpoint to get a transaction receipt returns `table_id`, when there is potentially many tables associated with a transaction.
We should consider changing the schema of the JSON data returned from /receipt/{chainId}/{transactionHash} to allow sending back all of the table IDs associated with the transaction.

Specifically this line would need to change from `type: string` to `type: array` with `items` as `type: string`.

[https://github.com/tablelandnetwork/docs/blob/main/specs/validator/tableland-openapi-spec.yaml#L248](https://github.com/tablelandnetwork/docs/blob/main/specs/validator/tableland-openapi-spec.yaml#L248)

The underlying system database table, `system_txn_receipts`, may or may not need to change.  We could potentially change the `table_id` column to accommodate a list of IDs, or perhaps we could insert a row for each unique combination of chain_id, txn_hash, and table_id.  This might already be how the table was designed?

I've marked this as a bug, but maybe it should be a feature request.  My thinking is that since we have been supporting transactions that support mutating more than one table since August 2022, we should have been returning all of the table IDs. However the only way to create this type of transaction was via direct contract to contract interaction with the Registry, and maybe aquiring all of the table IDs is a feature?

<sub>From [SyncLinear.com](https://synclinear.com) | [GOT-92](https://linear.app/tableland/issue/GOT-92/[got-93]-the-validator-api-endpoint-to-get-a-transaction-receipt)</sub>

brunocalza:
# Summary

- This PR implements the [new API spec (1.0.1)](https://github.com/tablelandnetwork/docs/pull/112).

# Context

Fixes #522 

# Implementation overview

- Most of the changes are to reflect the passage of the `TableIds` data through multiple layers
- Adds a new migration to store `table_ids` in the `system_txn_receipts` table
- It's hard to create an e2e test of this feature. I've added a test at the controller level with a mocked gateway.

### DIFF
--- a/cmd/api/main.go
+++ b/cmd/api/main.go
@@ -30,7 +30,6 @@ import (
 	"github.com/textileio/go-tableland/pkg/database"
 	"github.com/textileio/go-tableland/pkg/eventprocessor"
 	"github.com/textileio/go-tableland/pkg/eventprocessor/eventfeed"
-
 	"go.opentelemetry.io/otel/attribute"
 
 	efimpl "github.com/textileio/go-tableland/pkg/eventprocessor/eventfeed/impl"
@@ -72,16 +71,16 @@ func main() {
 		path.Join(dirPath, "database.db"),
 	)
 
-	db, err := database.Open(databaseURL, attribute.String("database", "main"))
-	if err != nil {
-		log.Fatal().Err(err).Msg("opening the read database")
-	}
-
 	// Restore provided backup (if configured).
 	if config.BootstrapBackupURL != "" {
 		if err := restoreBackup(databaseURL, config.BootstrapBackupURL); err != nil {
 			log.Fatal().Err(err).Msg("restoring backup")
 		}
+	}
+
+	db, err := database.Open(databaseURL, attribute.String("database", "main"))
+	if err != nil {
+		log.Fatal().Err(err).Msg("opening the read database")
 	}
 
 	// Parser.
--- a/internal/gateway/gateway.go
+++ b/internal/gateway/gateway.go
@@ -151,9 +151,12 @@ func (g *GatewayService) GetReceiptByTransactionHash(
 		BlockNumber:   receipt.BlockNumber,
 		IndexInBlock:  receipt.IndexInBlock,
 		TxnHash:       receipt.TxnHash,
-		TableID:       receipt.TableID,
+		TableIDs:      receipt.TableIDs,
 		Error:         receipt.Error,
 		ErrorEventIdx: receipt.ErrorEventIdx,
+
+		// Deprecated
+		TableID: receipt.TableID,
 	}, true, nil
 }
 
@@ -195,14 +198,17 @@ func (g *GatewayService) emptyMetadataImage() string {
 
 // Receipt represents a Tableland receipt.
 type Receipt struct {
-	ChainID      tableland.ChainID
-	BlockNumber  int64
-	IndexInBlock int64
-	TxnHash      string
-
-	TableID       *tables.TableID
+	ChainID       tableland.ChainID
+	BlockNumber   int64
+	IndexInBlock  int64
+	TxnHash       string
+	TableIDs      []tables.TableID
 	Error         *string
 	ErrorEventIdx *int
+
+	// Deprecated: the Receipt must hold information of all tables that were modified by the transaction.
+	// This field was replaced by TableIDs.
+	TableID *tables.TableID
 }
 
 // Table represents a system-wide table stored in Tableland.
--- a/internal/gateway/impl/gateway_store.go
+++ b/internal/gateway/impl/gateway_store.go
@@ -144,18 +144,33 @@ func (s *GatewayStore) GetReceipt(
 		IndexInBlock: res.IndexInBlock,
 		TxnHash:      txnHash,
 	}
+
 	if res.Error.Valid {
 		receipt.Error = &res.Error.String
 
 		errorEventIdx := int(res.ErrorEventIdx.Int64)
 		receipt.ErrorEventIdx = &errorEventIdx
 	}
+
 	if res.TableID.Valid {
 		id, err := tables.NewTableIDFromInt64(res.TableID.Int64)
 		if err != nil {
-			return gateway.Receipt{}, false, fmt.Errorf("parsing id to string: %s", err)
+			return gateway.Receipt{}, false, fmt.Errorf("parsing id integer: %s", err)
 		}
-		receipt.TableID = &id
+		receipt.TableID = &id // nolint
+	}
+
+	if res.TableIds.Valid {
+		tableIdsStr := strings.Split(res.TableIds.String, ",")
+		tableIds := make([]tables.TableID, len(tableIdsStr))
+		for i, idStr := range tableIdsStr {
+			tableID, err := tables.NewTableID(idStr)
+			if err != nil {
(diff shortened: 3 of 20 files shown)


Makefile
```diff
@@ -80,7 +80,7 @@ lint:
 .PHONY: lint
 
 # OpenAPI
-SPEC_URL=https://raw.githubusercontent.com/tablelandnetwork/docs/main/specs/validator/tableland-openapi-spec.yaml
+SPEC_URL=https://raw.githubusercontent.com/tablelandnetwork/docs/bcalza/fixreceipt/specs/validator/tableland-openapi-spec.yaml
```
brunocalza, line 83: This needs to be changed to `main` after the review

pkg/eventprocessor/impl/executor/impl/txnscope.go
```diff
@@ -107,9 +108,16 @@ func (ts *txnScope) executeTxnEvents(
 				ErrorEventIdx: &idx,
 			}, nil
 		}
+
+		if res.TableID != nil {
```
brunocalza, line 112: This is the core of the bug fix. Inside the loop of events that are being processed, we build a slice of `tableIds`. Before, we were simply getting the last table id, which is inside `res.TableID`

internal/gateway/impl/gateway_store.go
```diff
@@ -144,20 +144,35 @@ func (s *GatewayStore) GetReceipt(
...
 	if res.TableID.Valid {
 		id, err := tables.NewTableIDFromInt64(res.TableID.Int64)
 		if err != nil {
-			return gateway.Receipt{}, false, fmt.Errorf("parsing id to string: %s", err)
+			return gateway.Receipt{}, false, fmt.Errorf("parsing id integer: %s", err)
 		}
 		receipt.TableID = &id
 	}
 
+	if res.TableIds.Valid {
```
brunocalza, line 163: if the `TableIds` is not null, we split the string to get the table ids separately

internal/router/controllers/controller.go
```diff
@@ -138,6 +138,13 @@ func (c *Controller) GetReceiptByTransactionHash(rw http.ResponseWriter, r *http
 		receiptResponse.ErrorEventIdx = int32(*receipt.ErrorEventIdx)
 	}
 
+	ids := make([]string, len(receipt.TableIDs))
```
brunocalza, line 141: build an array of table ids and add to the response

pkg/database/migrations/005_receipttableids.up.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE system_txn_receipts ADD table_ids TEXT;
```
brunocalza, line 1: Migration that adds the `table_ids` columns. Also, we need to transfer the values in `table_id` to `table_ids`

pkg/eventprocessor/impl/executor/impl/blockscope.go
```diff
@@ -137,14 +137,23 @@ func (bs *blockScope) SaveTxnReceipts(ctx context.Context, rs []eventprocessor.R
 			tableID.Valid = true
 			tableID.Int64 = r.TableID.ToBigInt().Int64()
 		}
+
 		if r.Error != nil {
 			*r.Error = strings.ToValidUTF8(*r.Error, "")
 		}
+
+		tableIDs := sql.NullString{Valid: false}
```
brunocalza, line 145: adjust the `SaveTxnReceipts` method to store the table ids

cmd/api/main.go
```diff
@@ -72,18 +71,18 @@ func main() {
 		path.Join(dirPath, "database.db"),
 	)
 
-	db, err := database.Open(databaseURL, attribute.String("database", "main"))
```
brunocalza, line 75: We should open the database after `restoreBackup` because if the backup does not contain a migration the migration should be run. Otherwise, the migration would be run on an empty database and backup would overwrite that.

joewagner (approved):
This all looks good.  I setup a build of this branch with a local version of all of our npm packages and the new openapi spec doc, and did some testing.  Everything works as expected so I think we are good to go here.
We will have to coordinate releasing this, the docs change, and a small change to the js-sdk that is currently a draft PR https://github.com/tablelandnetwork/js-tableland/pull/479

brunocalza:
> This all looks good. I setup a build of this branch with a local version of all of our npm packages and the new openapi spec doc, and did some testing. Everything works as expected so I think we are good to go here. We will have to coordinate releasing this, the docs change, and a small change to the js-sdk that is currently a draft PR [tablelandnetwork/js-tableland#479](https://github.com/tablelandnetwork/js-tableland/pull/479)

Yeah. Let's aim to release it next week. how's that sound?

joewagner:
> Yeah. Let's aim to release it next week. how's that sound?
Sounds perfect.

cmd/api/main.go
avichalp: good catch 👍

pkg/database/migrations/005_receipttableids.up.sql
avichalp: Would it make sense to make `table_ids` "NOT NULL"? Would we always have a `table_id/table_ids` associated with the receipt that we are interested in saving?

internal/gateway/gateway.go
```diff
@@ -195,14 +198,16 @@ func (g *GatewayService) emptyMetadataImage() string {
...
+	ChainID       tableland.ChainID
+	BlockNumber   int64
+	IndexInBlock  int64
+	TxnHash       string
+	TableIDs      []tables.TableID
 	Error         *string
 	ErrorEventIdx *int
+
+	// Deprecated
+	TableID *tables.TableID
```
avichalp, line 210: maybe it would be good to add a comment about the reason for the deprecation and a link to issue #522

pkg/database/migrations/005_receipttableids.up.sql
avichalp: One issue I see with backfilling is that earlier, we were only saving the last `table_id`. This means that for these rows we will not have the correct value of `table_ids` (all tables associated with the txn). Caveat for future debugging.

pkg/tables/tables.go
```diff
@@ -4,13 +4,26 @@ import (
...
 )
 
 // TableID is the ID of a Table.
 type TableID big.Int
 
+// TableIDs is a list of TableID.
+type TableIDs []TableID
+
+// String transform a list of TableIds into a string.
+func (ids TableIDs) String() string {
```
avichalp, line 19: 👍

avichalp (dismissed):

pkg/database/migrations/005_receipttableids.up.sql
brunocalza: > Would it make sense to make table_ids "NOT NULL"? Would we always have a table_id/table_ids associated with the receipt that we are interested in saving?

`table_ids` can be null when an error occurs

> One issue I see with backfilling is that earlier, we were only saving the last table_id. This means that for these rows we will not have the correct value of table_ids (all tables associated with the txn). Caveat for future debugging.

That's fine because we shouldn't have any txns with more than one table id in `mainnet`

avichalp: got it, thanks.

PR #556, opened 2023-05-02
fix table's name validation

brunocalza:
# Summary

See https://github.com/tablelandnetwork/go-sqlparser/pull/54


joewagner:
@brunocalza I used a build of this and all the third party SDK integration tests pass, so I think this should work great.

avichalp (approved):

joewagner (approved):
LGTM!

PR #562, opened 2023-05-04
adds timeout to filter logs operation

brunocalza:
# Summary

We noticed that the WebSocket connection with Alchemy was hanging after getting a 429 error. Not really sure why that was happening. Maybe it's not the connection per se, but the Ethereum Client that is getting blocked. Not sure. I've add an explicit timeout to the filter logs call to see if fixes the problem


avichalp (approved):
LGTM

PR #564, opened 2023-05-09
removes healthbot probes for goerli chains

brunocalza:
# Summary

Removes Arbitrum Goerli and Optimism Goerli healthbot probes

# Context

It got too hard to fund those wallets. It's not making sense to have those probes.


avichalp (approved):
Nice!

PR #566, opened 2023-05-10
move the balance metrics to healthbot

brunocalza:
# Summary

This PR moves the wallet balance metrics to `healthbot`. 

# Context

We keep track of the wallet balances to know if `healthbot` wallets are running low. Historically, `healthbot` used to rely on relaying for emitting transactions. That meant that the validator had private keys necessary for relaying. The code that was doing the relay was keeping track of balances. When we removed relaying, we kept the balance tracking inside the validator unnecessarily. 

As a side effect, the validator required a private key even though it did not need it for its operations. And this additional configuration is a barrier to entry for new validators. 

With this PR, we can remove the section [Create and fund wallets](https://www.notion.so/textile/Validator-documentation-9f0cc2abf424410c8659fa939ed5095e?pvs=4#a32f368045754476be8c167ff922b9d4) from **Validator documentation**.

# Changes

- Removes the need to configure private keys for validators
- Changes Grafana dashboard so that `healthbot` metrics have their own row that can be toggled for `mainnet` to not appear
Screen Shot 2023-05-10 at 18 57 06


cmd/api/config.go
```diff
@@ -102,9 +102,6 @@ type ChainConfig struct {
 		EthEndpoint     string `default:"eth_endpoint"`
 		ContractAddress string `default:"contract_address"`
 	}
-	Signer struct {
```
brunocalza, line 105: no more need for private keys config

cmd/api/config.go
```diff
@@ -115,11 +112,6 @@ type ChainConfig struct {
 		BlockFailedExecutionBackoff string `default:"10s"`
 		DedupExecutedTxns           bool   `default:"false"`
 	}
-	NonceTracker struct {
```
brunocalza, line 118: the nonce tracker was the module that was keeping track of balances, we're not going to start it anymore. so we don't need these configs

cmd/api/main.go
```diff
@@ -166,48 +164,6 @@ func createChainIDStack(
 	tableConstraints TableConstraints,
 	fetchExtraBlockInfo bool,
 ) (chains.ChainStack, error) {
-	conn, err := ethclient.Dial(config.Registry.EthEndpoint)
-	if err != nil {
-		return chains.ChainStack{}, fmt.Errorf("failed to connect to ethereum endpoint: %s", err)
-	}
-
-	wallet, err := wallet.NewWallet(config.Signer.PrivateKey)
```
brunocalza, line 174: no more need for wallets in the validator

cmd/api/main.go
```diff
@@ -166,48 +164,6 @@ func createChainIDStack(
...
-	if err != nil {
-		return chains.ChainStack{}, fmt.Errorf("parsing nonce tracker check interval duration: %s", err)
-	}
-	stuckInterval, err := time.ParseDuration(config.NonceTracker.StuckInterval)
-	if err != nil {
-		return chains.ChainStack{}, fmt.Errorf("parsing nonce tracker stuck interval duration: %s", err)
-	}
-	ctxLocalTracker, clsLocalTracker := context.WithTimeout(context.Background(), time.Second*15)
-	defer clsLocalTracker()
-	tracker, err := nonceimpl.NewLocalTracker(
```
brunocalza, line 193: no need for a nonce manager because we're not relaying

cmd/api/main.go
```diff
@@ -229,6 +185,11 @@ func createChainIDStack(
 		return chains.ChainStack{}, fmt.Errorf("creating event feed store: %s", err)
 	}
 
+	conn, err := ethclient.Dial(config.Registry.EthEndpoint)
```
brunocalza, line 188: this is not new, it was just moved

cmd/api/main.go
```diff
@@ -249,6 +210,12 @@ func createChainIDStack(
 		eventprocessor.WithDedupExecutedTxns(config.EventProcessor.DedupExecutedTxns),
 		eventprocessor.WithHashCalcStep(config.HashCalculationStep),
 	}
+
```
brunocalza, line 213: this is not new, it was just moved

cmd/healthbot/balancetracker.go
```diff
@@ -0,0 +1,183 @@
...
+	"github.com/textileio/go-tableland/pkg/metrics"
+	"github.com/textileio/go-tableland/pkg/wallet"
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/metric"
+	"go.opentelemetry.io/otel/metric/global"
+	"go.opentelemetry.io/otel/metric/instrument"
+)
+
+// BalanceTracker tracks the balance of a given wallet and produces metrics.
+type BalanceTracker struct {
```
brunocalza, line 25: new component to be used inside `healthbot`. for each chain we create a BalanceTracker, that makes a `BalanceAt` call to Ethereum provider

cmd/healthbot/balancetracker.go
```diff
@@ -0,0 +1,183 @@
...
+
+	t.mu.Lock()
+	t.currWeiBalance = gWeiBalance
+	t.ethClientUnhealthy = 0
+	t.mu.Unlock()
+
+	return nil
+}
+
+func (t *BalanceTracker) initMetrics(chainID int, addr common.Address) error {
```
brunocalza, line 127: initialize otel metrics. we're using the same metric names we had before

docker/observability/grafana/provisioning/dashboards/validator-dashboard.json
```diff
@@ -333,7 +258,7 @@
             "uid": "P1809F7CD0C75ACF3"
           },
           "exemplar": true,
-          "expr": "sum by (chain_id) (tableland_wallettracker_balance_wei{service_name=\"tableland:api\"})/1000000000",
+          "expr": "sum by (chain_id) (tableland_wallettracker_balance_wei{service_name=\"tableland:healthbot\"})/1000000000",
           "format": "time_series",
```
avichalp, line 262: i am thinking if we could filter out the gorli based chains from here since we cannot do much to maintain their balance 🤔

avichalp (approved):
Looks good 👍

docker/observability/grafana/provisioning/dashboards/validator-dashboard.json
brunocalza: it's already filtered because the `healthbot` does not produce those metrics (see the image in the PR description)

avichalp: ah! got it. it is filtered because wallet metrics are now coming from healthbot not API 👍

PR #572, opened 2023-05-22
Implement Discord webhook for Tableland txns

avichalp:
# Summary

This pull request introduces a new feature to the EventProcessor component of the Tableland project. Specifically, it implements a Webhook module. If the webhook is `enabled` for a chain, The `EventProcessor` will send the update about the transaction to a Discord webhook endpoint.

# Implementation overview

Webhook Implementation adds support for dispatching processed events to a designated Discord webhook URL. It leverages Go's standard HTTP and JSON libraries to send POST requests with appropriate payloads. Webhook Interface and Discord Webhook Struct provide an abstraction layer to allow future expansion of the webhook module with additional services such as IFTTT etc. For now, only Discord is supported.

# Checklist

- [ ]  Are changes backward compatible with existing SDKs, or is there a plan to manage it correctly?
- [ ]  Are code changes optimized for future code readers, commenting on problematic areas to understand (if any)?
- [ ]  Future-self question: Did you avoid unjustified/unnecessary complexity to achieve the goal?
- [ ] Create environment vars in mainnet for `VALIDATOR_DISCORD_WEBHOOK_ID` and `VALIDATOR_DISCORD_WEBHOOK_TOKEN`


cmd/api/config.go
```diff
@@ -53,6 +53,13 @@ type HTTPConfig struct {
 	MaxRequestPerInterval uint64 `default:"10"`
 }
 
+// WebhookConfig contains configuration for the Table events Webhook.
+type WebhookConfig struct {
```
avichalp, line 57: New webhook config is introduced at the `EventProcessor` level.

cmd/api/main.go
```diff
@@ -211,6 +211,14 @@ func createChainIDStack(
 		eventprocessor.WithHashCalcStep(config.HashCalculationStep),
 	}
 
+	// Add the webhook config if it is enabled for this chain.
+	if config.EventProcessor.Webhook.Enabled {
```
avichalp, line 215: If the chain has webhooks enabled, we configure the `EventProcessor` accordingly.

pkg/eventprocessor/impl/eventprocessor.go
```diff
@@ -296,6 +313,43 @@ func (ep *EventProcessor) executeBlock(ctx context.Context, block eventfeed.Bloc
 	return nil
 }
 
+// executeWebhook will iterate over the receipts and send a webhook for each
+// receipt. We do this in a separate goroutine to avoid blocking.
+func (ep *EventProcessor) executeWebhook(ctx context.Context, receipts []eventprocessor.Receipt) {
+	for _, r := range receipts {
```
avichalp, line 319: For each receipt, we spin up a new goroutine to make the webhook call

pkg/eventprocessor/impl/webhook.go
```diff
@@ -0,0 +1,83 @@
...
+
+// Send method for DiscordWebhook.
+func (w *DiscordWebhook) Send(ctx context.Context, content string) error {
+	return sendWebhookRequest(ctx, w.URL, webhookContent{
+		Content: content,
+	})
+}
+
+// NewWebhook function to create a new webhook.
+func NewWebhook(endpointType string, url string) (Webhook, error) {
```
brunocalza, line 75: maybe adding an additional check just to see if `url` is valid? I think the method is `url.Parse`

pkg/eventprocessor/impl/eventprocessor.go
```diff
@@ -296,6 +313,43 @@ func (ep *EventProcessor) executeBlock(ctx context.Context, block eventfeed.Bloc
...
+		go func(r eventprocessor.Receipt) {
+			successTmpl := "**Error processing Tableland event:**\n\n" +
```
brunocalza, line 321: suggestions:
- Put this logic inside the Webhook object. Right now the Webhook is just responsible for sending a `content` string, and the responsibility of creating the content is `eventprocessor`. Not sure `eventprocessor` should be responsible for that. That would imply passing the Receipt to `Send`. That change changes the meaning of Webhook from a simply HTTP content forwarder to a "Receipt Webhook" forwarder. 
- Maybe using `text/template` for producing the content, instead of `Sprinft`?

pkg/eventprocessor/impl/webhook_test.go
```diff
@@ -0,0 +1,194 @@
...
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/textileio/go-tableland/pkg/eventprocessor"
+	"github.com/textileio/go-tableland/pkg/tables"
+)
+
+func TestExecuteWebhook(t *testing.T) {
```
brunocalza, line 16: nice test :)

pkg/eventprocessor/impl/webhook.go
```diff
@@ -0,0 +1,83 @@
...
+
+	if resp.StatusCode >= 400 {
+		return fmt.Errorf("webhook request failed with status code: %d", resp.StatusCode)
+	}
+
+	return nil
+}
+
+// Webhook interface for sending webhooks to different services such as IFTTT or Discord etc.
+type Webhook interface {
```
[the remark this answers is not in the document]
brunocalza: Considering my comment about having just the URL we would not need different implementations. In fact, if you look at the `DiscordWebhook` implementation 

```
func (w *DiscordWebhook) Send(ctx context.Context, content string) error {
	return sendWebhookRequest(ctx, w.URL, webhookContent{
		Content: content,
	})
}
```

there's nothing in there that makes it specific to Discord, because all the Discord information is already encoded in the URL. So if you were to create another implementation, let's say Slack, it would be literally the same implementation:

```
func (w *SlackWebhook) Send(ctx context.Context, content string) error {
	return sendWebhookRequest(ctx, w.URL, webhookContent{
		Content: content,
	})
}
```

In a way, that is evidence that we don't need `EndpointType`. 

Just for completeness. It doesn't mean we don't need to treat different endpoints differently. It could be that we could need logic in the implementation to deal if the specifics of which provider. Maybe retrying logic, rate limiting handling, or error handling. And that scenario could lead to the need for different implementations. But I don't think that's the case now

brunocalza (commented):
Looks good! Wrote some suggestions for your consideration. Let me know what you think

cmd/api/config.go
avichalp: The first two points make sense. Removing the `enabled` flag and adding a check to see if the URL is not an empty string. Still in doubt about the third point. Please see the other comment.

pkg/eventprocessor/impl/webhook.go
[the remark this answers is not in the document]
brunocalza: Got it! yeah, I did not realize the requests could be different and the specifics of Discord were in the `Content` fields. We would definitely need different implementations. Sounds good then to keep it. The interface you'd still need for testing.

cmd/api/config.go
brunocalza: > Still in doubt about the third point. Please see the other comment.

It's not strictly needed. As you said, we could parse the URL to decide which implementation to use. I'm fine either way.

pkg/eventprocessor/impl/webhook.go
```diff
@@ -0,0 +1,135 @@
...
+	"net/http"
+	"net/url"
+	"time"
+
+	logger "github.com/rs/zerolog/log"
+	"github.com/textileio/go-tableland/pkg/eventprocessor"
+)
+
+// webhookTemplate is the template used to generate the webhook content.
+const contentTemplate = `
```
brunocalza, line 18: looks nice!

pkg/eventprocessor/impl/eventprocessor.go
```diff
@@ -87,6 +89,15 @@ func New(
 		return nil, fmt.Errorf("initializing metric instruments: %s", err)
 	}
 
+	if config.WebhookURL != "" {
+		whe, err := NewWebhook(config.WebhookURL)
+		if err != nil {
+			return nil, fmt.Errorf("webhook endpoint cannot be initialized: %s", err)
```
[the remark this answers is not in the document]
avichalp: so, these are the cases when there is a URL configured, but it is "invalid" (unexpected hostname, malformed URL, etc). 

i think there are two cases. one, where the validator had accidentally put in the wrong URL. In that case, they might want to know immediately, so "silent failing" might not be good. in the other case, they might be trying to deliberately put in a wrong value, but i am not sure why they would do that. they could just remove the URL from the config. so, i think it might be okay not to let them start the validator and force them to fix the URL. What do you think?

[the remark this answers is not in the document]
brunocalza: sounds good! The validator will be initialized, it's just the event processor for that chain with misconfigured URL that will not

brunocalza (dismissed):

pkg/eventprocessor/impl/eventprocessor.go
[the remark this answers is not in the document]
avichalp: yes, i meant the validator will start, but then it will `exit` when it tries to configure the `chainStack` [here](https://github.com/tablelandnetwork/go-tableland/pull/572/files#diff-ae22ad1e20b89eaefa8e09e4ed89443ac10fdc555b9767625f3866e3853100afR229), right?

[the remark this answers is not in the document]
brunocalza: oh yeah, you're right!

brunocalza (approved):

PR #578, opened 2023-05-23
fixes table ids duplication in receipts

brunocalza:
# Summary

The `table_ids` were duplicated in the receipt in the case of multiple events from the same table. 
This PR fixes code to avoid that but does not fix the data. That will be done with an update directly in the database.

### DIFF
--- a/pkg/eventprocessor/impl/executor/impl/txnscope.go
+++ b/pkg/eventprocessor/impl/executor/impl/txnscope.go
@@ -56,7 +56,7 @@ func (ts *txnScope) executeTxnEvents(
 	var res eventExecutionResult
 	var err error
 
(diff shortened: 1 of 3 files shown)


pkg/eventprocessor/impl/executor/impl/txnscope.go
```diff
@@ -56,7 +56,7 @@ func (ts *txnScope) executeTxnEvents(
 	var res eventExecutionResult
 	var err error
 
-	tableIDs := make([]tables.TableID, 0)
+	tableIDs, tableIDsMap := make([]tables.TableID, 0), make(map[string]struct{})
```
brunocalza, line 59: adds a map to avoid duplicates

PR #580, opened 2023-05-26
Replace Hyperspace with Calibration

avichalp:
# Summary

Fixes GOT-110

This PR replaces Hyperspace with Calibration testnet. The hyperspace network will be shut down at the end of May. Since Ankr doesn't support Calibration yet, we must use Glif.io. 

Currently, Glif doesn't have the historical data because of an issue in Lotus. But this issue will be resolved in the coming weeks. Until then we can work with Calibration by ignoring the older events. It is not ideal, but as a testnet, it can be used to make demos or organize hackathons. Once the Archive node is available on Glif.io, we can use a custom node with authentication. 

This PR also removes Hyperspace from the `healthbot` but doesn't add calibration yet. We have to manually deploy the `healthbot` table because the one that is automatically created by `evm-tableland` will not be replayed on the validator due to historical data constraints. 

I will add the `healthbot` configuration in a follow-up PR after creating the table manually. 

- [x] Add env variable `VALIDATOR_GLIF_FILECOIN_CALIBRATION_API_KEY` on testnet


pkg/eventprocessor/eventfeed/impl/eventfeed.go
```diff
@@ -184,6 +184,20 @@ func (ef *EventFeed) Start(
 					strings.Contains(err.Error(), "block range is too wide") {
 					ef.maxBlocksFetchSize = ef.maxBlocksFetchSize * 80 / 100
 				} else {
+					// If we get a "looksback" error it means that history is not available
+					// for this chain. It happens in Filecoin based chains, where the
+					// history is not available in non archive nodes.
+					// In this case, we just move the fromHeight
+					// to the current head minus 1995 blocks and ignore the past events.
+					// This is temporary until we have a better access to archive nodes.
+					if strings.Contains(err.Error(), "lookbacks of more than") {
```
avichalp, line 193: Ignoring events deeper than 1995 blocks because non archival nodes don't have access to it.

cmd/healthbot/main.go
```diff
@@ -58,10 +58,11 @@ func main() {
...
 		}
 
-		// For Filecoin Hyperspace, we use Ankr endpoint
+		// For Filecoin Calibration, we use Glif endpoint
 		opts := []clientV1.NewClientOption{clientV1.NewClientChain(chain)}
-		if chain.ID == 3141 {
-			opts = append(opts, clientV1.NewClientAnkrAPIKey(chainCfg.AnkrAPIKey))
+		if chain.ID == 314159 {
+			// Glif API key is empty string because currently we are using the free tier (public node)
+			opts = append(opts, clientV1.NewClientGlifAPIKey(""))
```
avichalp, line 65: Currently API key is empty since we are using the public node.

brunocalza (approved):
LGTM!

brunocalza (approved):

PR #589, opened 2023-07-05
Use Glif API for Filecoin mainnet

avichalp:
This PR replaces Anrk's Filecoin API with Glif's for FEVM integration.  Earlier, we used Ankr APIs, but we had to move to Glifio because Ankr doesn't give access to the Filecoin archive node. Changes have been made to the configuration parameters in order to support the Glif API auth token. An additional check is introduced for the Filecoin chain to include this auth token in the request header.

Fixes: GOT-97

- [x] Add env var to mainnet VALIDATOR_GLIF_FILECOIN_MAINNET_API_KEY
- [ ] ~Rewind the `system_evm_blocks` table for Filecoin mainnet~ (since only healthbot table will be affected, we decided to not sync Filecoin mainnet from scratch)


cmd/api/main.go
```diff
@@ -185,11 +186,20 @@ func createChainIDStack(
...
 
-	conn, err := ethclient.Dial(config.Registry.EthEndpoint)
+	ethRPCClient, err := ethrpc.Dial(config.Registry.EthEndpoint)
 	if err != nil {
 		return chains.ChainStack{}, fmt.Errorf("failed to connect to ethereum endpoint: %s", err)
 	}
 
+	// For the Filecoin (314) chain, we need to set the auth token
+	// in the header of the request.
+	if config.ChainID == 314 {
```
brunocalza, line 196: What about checking if `ProviderAuthToken` is not empty?

docker/deployed/mainnet/api/config.json
```diff
@@ -155,8 +155,9 @@
             "Name": "Filecoin Mainnet",
             "ChainID": 314,
             "Registry": {
-                "EthEndpoint": "https://rpc.ankr.com/filecoin/${VALIDATOR_ANKR_FILECOIN_MAINNET_API_KEY}",
-                "ContractAddress": "0x59EF8Bf2d6c102B4c42AEf9189e1a9F0ABfD652d"
+                "EthEndpoint": "https://node.glif.io/fvm-archive/lotus/rpc/v1",
+                "ContractAddress": "0x59EF8Bf2d6c102B4c42AEf9189e1a9F0ABfD652d",
+                "ProviderAuthToken": "${VALIDATOR_GLIF_FILECOIN_MAINNET_API_KEY}"
             },
```
avichalp, line 161: Adds Glif token in the config.

brunocalza (dismissed):
Looks good to me 👍

brunocalza (approved):

PR #597, opened 2023-07-13
Reformat webhooks

avichalp:
### Summary

Adds more context to webhooks content by adding hyperlinks for transactions, table ids and chain ids. Fixes GOT-108

### Implementation overview

To add links to the tableland docs, and block explorers I have stored these links in the webhooks package against their corresponding chain ids. 

### DIFF
--- a/pkg/eventprocessor/impl/webhook.go
+++ b/pkg/eventprocessor/impl/webhook.go
@@ -89,98 +89,86 @@ type chain struct {
 	SupportsNFTView  bool
 }
 
-var chains = map[int64]chain{
+var chains = map[tableland.ChainID]chain{
+	// Ethereum
 	1: {
-		ID:               1,
-		Name:             "Ethereum",
 		ContractAddr:     common.HexToAddress("0x012969f7e3439a9B04025b5a049EB9BAD82A8C12"),
 		TBLDocsURL:       "https://docs.tableland.xyz/fundamentals/chains/ethereum",
 		BlockExplorerURL: "https://etherscan.io",
 		SupportsNFTView:  true,
 	},
+	// Optimism
 	10: {
-		ID:               10,
-		Name:             "Optimism",
 		ContractAddr:     common.HexToAddress("0xfad44BF5B843dE943a09D4f3E84949A11d3aa3e6"),
 		TBLDocsURL:       "https://docs.tableland.xyz/fundamentals/chains/optimism",
 		BlockExplorerURL: "https://optimistic.etherscan.io",
 		SupportsNFTView:  false,
 	},
+	// Polygon
 	137: {
-		ID:               137,
-		Name:             "Polygon",
 		ContractAddr:     common.HexToAddress("0x5c4e6A9e5C1e1BF445A062006faF19EA6c49aFeA"),
 		TBLDocsURL:       "https://docs.tableland.xyz/fundamentals/chains/polygon",
 		BlockExplorerURL: "https://polygonscan.com",
 		SupportsNFTView:  false,
 	},
+	// Arbitrum One
 	42161: {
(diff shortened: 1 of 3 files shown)


pkg/eventprocessor/impl/webhook.go
```diff
@@ -76,14 +80,163 @@ func sendWebhookRequest(ctx context.Context, url string, body interface{}) error
 	return nil
 }
 
+type chain struct {
+	ID               int64
+	Name             string
+	ContractAddr     common.Address
+	TBLDocsURL       string
+	BlockExplorerURL string
+	SupportsNFTView  bool
```
avichalp, line 89: Some block explorers like Etherscan have started to provide an NFT view. I believe this feature will be added to others eventually. Using this flag we can provide a link to the NFT view if available.

pkg/eventprocessor/impl/webhook.go
```diff
@@ -76,14 +80,162 @@ func sendWebhookRequest(ctx context.Context, url string, body interface{}) error
 	return nil
 }
 
+type chain struct {
+	ID               int64
```
brunocalza, line 84: Is the ID field necessary? I'm not seeing it being used and also it seems a bit redundant because the key of the map is the chain ID

brunocalza: I think Name is also not being used

pkg/eventprocessor/impl/webhook.go
```diff
@@ -76,14 +80,162 @@ func sendWebhookRequest(ctx context.Context, url string, body interface{}) error
...
+	ContractAddr     common.Address
+	TBLDocsURL       string
+	BlockExplorerURL string
+	SupportsNFTView  bool
+}
+
+var chains = map[int64]chain{
+	1: {
+		ID:               1,
+		Name:             "Ethereum",
```
brunocalza, line 95: If you remove the Name field, you can add a comment on top of each chain to indicate the name of the chain

pkg/eventprocessor/impl/webhook.go
```diff
@@ -76,14 +80,162 @@ func sendWebhookRequest(ctx context.Context, url string, body interface{}) error
...
+	TableIDs      string
+	Error         *string
+	ErrorEventIdx *int
+}
+
+// getTableNFTViews returns the NFT views for the given table IDs.
+func getTableNFTViews(tableIDs tables.TableIDs, chainID tableland.ChainID) string {
+	var tableNFTURLs []string
+	for _, tableID := range tableIDs {
+		if !chains[int64(chainID)].SupportsNFTView {
```
brunocalza, line 206: worth creating a variable for storing the value of `chains[int64(chainID)]`, so you don't do lots map lookup later? mostly to increase readability

pkg/eventprocessor/impl/webhook.go
```diff
@@ -76,14 +80,162 @@ func sendWebhookRequest(ctx context.Context, url string, body interface{}) error
...
+type chain struct {
+	ID               int64
+	Name             string
+	ContractAddr     common.Address
+	TBLDocsURL       string
+	BlockExplorerURL string
+	SupportsNFTView  bool
+}
+
+var chains = map[int64]chain{
```
brunocalza, line 92: What happens if you change to `map[tableland.ChainID]chain`? does it avoid the `int64` conversions?

brunocalza (approved):
Looks good! Left some comments for consideration

pkg/eventprocessor/impl/webhook.go
avichalp: Yea, good idea. Its not being used anywhere.

PR #599, opened 2023-07-25
allows authorized clients to bypass rate limiter

brunocalza:
# Summary

This PR adds a new config for the HTTP rate limiter: `ApiKey`. With that, authorized clients (clients that provide the same key) will not be rate limited. 

# Context

Our Studio application is being rate limited fairly easily. 

# Implementation overview

Adds a new config to be configured via env. Makes a change to the rate limiter middleware to ignore the authorized clients

### DIFF
--- a/cmd/api/config.go
+++ b/cmd/api/config.go
@@ -51,7 +51,7 @@ type HTTPConfig struct {
 
 	RateLimInterval       string `default:"1s"`
 	MaxRequestPerInterval uint64 `default:"10"`
-	AllowList             string `default:""`
+	AllowList             string `default:""` // separated list of IPs (e.g. 127.0.0.1,192.168.0.1)
 }
 
 // GatewayConfig contains configuration for the Gateway.
--- a/cmd/api/config.go
+++ b/cmd/api/config.go
@@ -51,7 +51,7 @@ type HTTPConfig struct {
 
 	RateLimInterval       string `default:"1s"`
 	MaxRequestPerInterval uint64 `default:"10"`
-	AllowList             string `default:""` // separated list of IPs (e.g. 127.0.0.1,192.168.0.1)
+	APIKey                string `default:""` // if client passes the key it will not be affected by rate limiter
 }
 
 // GatewayConfig contains configuration for the Gateway.
--- a/cmd/api/config.go
+++ b/cmd/api/config.go
@@ -51,6 +51,7 @@ type HTTPConfig struct {
 
 	RateLimInterval       string `default:"1s"`
 	MaxRequestPerInterval uint64 `default:"10"`
+	AllowList             string `default:""`
 }
 
 // GatewayConfig contains configuration for the Gateway.
--- a/cmd/api/main.go
+++ b/cmd/api/main.go
@@ -460,14 +460,12 @@ func createAPIServer(
 		return nil, fmt.Errorf("parsing http ratelimiter interval: %s", err)
 	}
 
-	allowList := strings.Split(httpConfig.AllowList, ",")
-
 	router, err := router.ConfiguredRouter(
 		g,
 		httpConfig.MaxRequestPerInterval,
 		rateLimInterval,
 		supportedChainIDs,
-		allowList,
+		httpConfig.APIKey,
 	)
 	if err != nil {
 		return nil, fmt.Errorf("configuring router: %s", err)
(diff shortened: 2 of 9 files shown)


internal/router/middlewares/ratelim_test.go
```diff
@@ -28,6 +29,9 @@ func TestLimit1IP(t *testing.T) {
 
 		{name: "success", callRPS: 100, limitRPS: 500, forwardedFor: false},
 		{name: "block-me", callRPS: 1000, limitRPS: 500, forwardedFor: false},
+
+		{name: "allow-me", callRPS: 1000, limitRPS: 500, forwardedFor: false, allow: true},
```
brunocalza, line 33: adds two more test cases where rps is greater than limit but it never gets 429

internal/router/middlewares/ratelim.go
```diff
@@ -77,3 +81,62 @@ func extractClientIP(r *http.Request) (string, error) {
 	}
 	return ip, nil
 }
+
+type middleware struct {
```
brunocalza, line 85: implementation borrowed from https://github.com/sethvargo/go-limiter/blob/main/httplimit/middleware.go with an extra tweak

internal/router/middlewares/ratelim.go
```diff
@@ -77,3 +81,62 @@ func extractClientIP(r *http.Request) (string, error) {
...
+		ctx := r.Context()
+
+		// Call the key function - if this fails, it's an internal server error.
+		key, err := m.keyFunc(r)
+		if err != nil {
+			http.Error(w, http.StatusText(http.StatusInternalServerError), http.StatusInternalServerError)
+			return
+		}
+
+		// skip rate limiting checks if key is in allowlist
```
brunocalza, line 109: this is the additional change

cmd/api/main.go
```diff
@@ -450,11 +450,14 @@ func createAPIServer(
 		return nil, fmt.Errorf("parsing http ratelimiter interval: %s", err)
 	}
 
+	allowList := strings.Split(httpConfig.AllowList, ",")
```
brunocalza, line 453: with this we can provide a list of addresses

internal/router/middlewares/ratelim.go
```diff
@@ -20,8 +22,9 @@ type RateLimiterConfig struct {
 // RateLimiterRouteConfig specifies the maximum request per interval, and
 // interval length for a rate limiting rule.
 type RateLimiterRouteConfig struct {
-	MaxRPI   uint64
-	Interval time.Duration
+	MaxRPI    uint64
+	Interval  time.Duration
+	AllowList []string
```
avichalp, line 27: 👍

cmd/api/config.go
```diff
@@ -51,6 +51,7 @@ type HTTPConfig struct {
 
 	RateLimInterval       string `default:"1s"`
 	MaxRequestPerInterval uint64 `default:"10"`
+	AllowList             string `default:""`
```
avichalp, line 54: maybe a comment with an example of a list of IP addrs would be helpful here.

internal/router/middlewares/ratelim_test.go
avichalp: looks good

avichalp (approved):
LGTM!

cmd/api/config.go
brunocalza: good idea 👍

internal/router/middlewares/ratelim.go
```diff
@@ -77,3 +81,62 @@ func extractClientIP(r *http.Request) (string, error) {
...
+
+		// Call the key function - if this fails, it's an internal server error.
+		key, err := m.keyFunc(r)
+		if err != nil {
+			http.Error(w, http.StatusText(http.StatusInternalServerError), http.StatusInternalServerError)
+			return
+		}
+
+		// skip rate limiting checks if secret key is provided
+		if key := r.Header.Get("Secret-Key"); key != "" && m.apiKey != "" {
```
avichalp, line 110: should we name the header "Api-Key" instead of "Secret-Key" to make it consistent with the code?

avichalp: hmm maybe i get your point now. this is the special "API key" that can bypass the checks. so it makes sense.

avichalp (approved):
Looks good!

PR #607, opened 2023-11-14
Migrate Arbitrum Network from Goerli to Sepolia

avichalp:
# Summary

This pull request includes changes to migrate the Arbitrum network from Goerli to Sepolia.

Key changes include:

**Update ChainIDs**: The ChainID for Arbitrum has been updated from ArbitrumGoerli to ArbitrumSepolia in the chains.go file. The corresponding ChainID number has also been updated.

**Update Configurations**: The .env_validator.example and config.json files have been updated to reflect the change in the Arbitrum network. The API key and endpoint for Arbitrum have been updated from Goerli to Sepolia.

**Update Endpoints**: The endpoint URLs for Arbitrum in the chains.go file have been updated to point to the Sepolia network.

**Update Contract Address**: The contract address for the Arbitrum network in the chains.go file has been updated.

These changes ensure that the application is now interacting with the Sepolia network for Arbitrum instead of Goerli.

- [x]  Are changes backward compatible with existing SDKs, or is there a plan to manage it correctly?
- [x]  Are changes covered by existing tests, or were new tests included?
- [x]  Are code changes optimized for future code readers, commenting on problematic areas to understand (if any)?
- [x]  Future-self question: Did you avoid unjustified/unnecessary complexity to achieve the goal?
- [x] Add required ENV vars

_NOTE: Doesn't add the healthbot config since we have stopped using healthbot on optimism and arbitrum_

PR #608, opened 2023-11-16
remove arb and eth goerli networks

avichalp:
# Summary

This PR removes support for arbitrum and ethereum goerli networks. See #607 for details


docker/deployed/testnet/api/config.json
```diff
@@ -52,25 +52,6 @@
         "ChainStackCollectFrequency": "15m"
     },
     "Chains": [
-        {
-            "Name": "Ethereum Goerli",
```
avichalp, line 56: cc: @dtbuchholz

brunocalza (approved):

asutula:
Will the validator still support read queries (and other types of gateway requests) for tables on these goerli networks?

avichalp:
@asutula i don't think so. the tables will still be there in the validator's database but the gateway has this notion of "supported chains". it rejects any query to an unsupported chain

brunocalza:
> Will the validator still support read queries (and other types of gateway requests) for tables on these goerli networks?

The query will work, actually. The check of supported chains does not happen in the `/query` call. Can't remember if that was by design or an accident. 

Either way, an additional comment: if we ever start the validator from scratch again, the data would be lost because the validator wouldn't pick the events of a not-supported chain (we'd still have them in the backups taken when the chain was supported). We don't have the concept of "network was supported between this interval, so pick those events up".

asutula:
Thanks for the info @brunocalza.

> if we ever start the validator from scratch again

This implies that if any new validator joins the nework, they won't have the data. Seems we should consider the data gone/unavailable then. I could be smart to delete the tables from the validator state (and Studio's state), but that may be another conversation.

asutula (approved):

PR #625, opened 2024-05-23
adds support for query parameters

brunocalza:
# Summary

- This PR adds support for the new SQL parser that is able to interpret placeholders (`?`) and replace them with query parameters that come from the request. You can now perform the following queries:
  
  - `GET /query?statement="select * from my_table where id = ?"&params=1`
  - `POST --json '{"statement": "SELECT * FROM my_table WHERE counter = ?", "params" : [1], "unwrap": true, "extract": true}'  /query `

- If you want to provide multiple params values, you just repeat the `params` with new values. For example `&params=1&params=2&params=3`. And if you want to provide a string, you gotta enclose it `&params="3"`.

- This PR depends on tablelandnetwork/go-sqlparser#388 and tablelandnetwork/docs#198

# Context

The context for this PR can be found at [ENG-852](https://linear.app/tableland/issue/ENG-852/dimo-request-add-placeholders-and-query-params-for-query-endpoint).

Make sure you understand tablelandnetwork/go-sqlparser#388 and tablelandnetwork/docs#198 before reviewing this one.

# Implementation overview 
This PR changes a lot of files, but most of the changes are just a side effect of the main changes. Here's an overview on the most important changes:

## Controller

The controller needed to be changed to capture the `params` query parameter. Both GET /query and POST /query were changed. The `params` is captured as a slice of string (`[]string`). In the case of the POST, since we have an `[]interface{}` after the JSON is decoded into memory, we try to convert it to `[]string`. 

## Resolver is now passed to the `GatewayService`

I made a change to make the resolver a dependency of the `GatewayService`. See the implication of this change [here](https://github.com/tablelandnetwork/go-tableland/pull/625/files#diff-ae22ad1e20b89eaefa8e09e4ed89443ac10fdc555b9767625f3866e3853100afR445-R450), [here](https://github.com/tablelandnetwork/go-tableland/pull/625/files#diff-68a1cc0c673b754f20006fd3916c6db4120489a84943d561de2a79507ef7f756L61-L93) and [here](https://github.com/tablelandnetwork/go-tableland/pull/625/files#diff-68a1cc0c673b754f20006fd3916c6db4120489a84943d561de2a79507ef7f756R169-R179). The reason for that is that it makes more sense for the `GatewayService` to attach the parameters values to the resolver, than the storage layer. The storage layer is just responsible for executing the query.

## The resolver is not a dependency of `GatewayStore` and is now passed in the `Read` method

We already mentioned how the resolver is passed to a `GatewayService` instead of `GatewayStore`. In order to build the final statement, we need to pass the resolver to the `Read` method. See it [here](https://github.com/tablelandnetwork/go-tableland/pull/625/files#diff-99f3427d646aa36492629d1d416848f95f0a0b4f08a08da61876e5a3cd901a55R32-R35):

## The resolver implementation needs to implement `GetBindValues() []sqlparser.Expr`

We must implement the new method of our resolver (remember [this](https://github.com/tablelandnetwork/go-sqlparser/pull/388/files#diff-0945956cd950d047e7fa866dcf34fee7d29cbce277d4e5e5cd621f2213f3be6cR124)? ). For that, we make use of [`PrepareParams`](https://github.com/tablelandnetwork/go-tableland/pull/625/files#diff-10e8c04527340840b0c7c8372c7fecd35c48df30792992a89f371588a3aa1e63R35) that converts `[]string` that comes from the controller to `[]sqlparser.Expr` and store those values inside the struct. [Here](https://github.com/tablelandnetwork/go-tableland/pull/625/files#diff-68a1cc0c673b754f20006fd3916c6db4120489a84943d561de2a79507ef7f756R175)'s how `PrepareParams` is used.

# Review orientation

1. Start at `internal/router/controllers/controller.go` and understand the change made in the controller
2. Go to `cmd/api/main.go` and see how the resolver is now passed to the gateway and not to the storage layer then check the changes at `internal/gateway/gateway.go`
3. Check the `internal/gateway/impl/gateway_store.go` file
4. Check the `pkg/parsing/readstatementresolver.go` file
5. Check the changes made in the Go client (`pkg/client/v1/readquery.go`)
6. The changes in the other files are just to adjust to the changes explained above


avichalp (dismissed):
looks good!

asutula (dismissed):

brunocalza:
@avichalp would you mind approving this again? i've updated with the correct parser version

avichalp (approved):

